PR #26, opened 2015-01-15
Resolve bug about expected order of parameter when do lambda.Invoke call 

Alex141:
Hello David!

A month ago I started using your library for my new project because is hasn't bugs that were in the original implementation of Microsoft (e.g. parsing of real numbers depends on the settings of the culture). 

Recently I found one bug which I want to notify you. The bug is as follows: order of params, that transfer to Lambda.Invoke() method should be the same as order of parameters, transfered to Interpreter.Parse() method, but is the same as the order in which the parameters are occured in the expression.
See example:

```C#
        var Parameter1 = new Parameter("a", typeof(int));
        var Parameter2 = new Parameter("b", typeof(int));

        var expr = new Interpreter().Parse("b-a", Parameter1, Parameter2);

        Console.WriteLine(expr.Invoke(1, 2));
        Console.ReadLine();
```

Expected: 1 (as a = 1, b = 2)
Real: 1 (b = 1, a = 2)

I resolved this bug and wrote test to reproduce it. I would like you accepted my pull request that resolve it and create new nuget packet  because my project CalcBinding uses your library quite so. Thank you!

### DIFF
--- a/src/DynamicExpresso.Core/ParserArguments.cs
+++ b/src/DynamicExpresso.Core/ParserArguments.cs
@@ -13,7 +13,7 @@ namespace DynamicExpresso
 		ParserSettings _settings;
 		Dictionary<string, Parameter> _parameters;
 
-		HashSet<Parameter> _usedParameters = new HashSet<Parameter>();
+		List<Parameter> _usedParameters = new List<Parameter>();
 		HashSet<ReferenceType> _usedTypes = new HashSet<ReferenceType>();
 		HashSet<Identifier> _usedIdentifiers = new HashSet<Identifier>();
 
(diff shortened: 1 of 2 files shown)


davideicardi:
Thank you very much! Good fixing. I just merged your changes and I will release a new version soon.

Alex141:
Thank you too!

davideicardi:
Due to some changes that I have done just today (for another bug) I modified slightly your code.
Basically instead of sort UsedParameters during parse I sort it during Invoke using the new DeclaredParameters property.

See https://github.com/davideicardi/DynamicExpresso/commit/56f81e887c54390d1664f7a1c502fc729a130bd4

Do you see any problem? I think that the idea is the same.

PR #30, opened 2015-03-03
ResolveExpression Callback

AdamsLair:
I have added an optional event to the Interpreter class that allows users to resolve a Property-, Field-, Method- or Extension Method Expression manually whenever the default resolve fails. A quick example:

```C#
public class Program
{
	public static void Main(string[] args)
	{
		// Trying to access the length of an object - this will not work without this update!
		Func<object,object> func;
		string expression = "arg.Length";

		// Creating an interpreter and subscribing to the optional resolve event
		Interpreter interpreter = new Interpreter();
		interpreter.ResolveUnknownExpression += ResolveUnknownExpression;
		
		// Now let's try to parse this...
		func = interpreter.ParseAsDelegate<Func<object,object>>(expression, "arg");

		// Works!
		Console.WriteLine(func("Hello"));
		Console.ReadLine();
	}

	// This event handler is called whenever the Interpreter fails to resolve an Expression.
	// It can be used to provide a suitable replacement on-the-fly. It will not be called when
	// the Parser is able to resolve the Expression by itself.
	private static void ResolveUnknownExpression(object sender, ResolveExpressionEventArgs e)
	{
		// If we're trying to resolve a Property or Field named "Length"...
		if (e.EventType == ResolveExpressionType.PropertyOrField && 
			e.Name == "Length")
		{
			// Let's just assume we're dealing with a string and cast it.
			e.Resolve(Expression.Property(Expression.TypeAs(e.Instance, typeof(string)), "Length"));
		}
	}
}
```

Please note that this is just a random example in order to illustrate how it works.

I will be using this for interpreting Expressions like `something.Parent.Name` where there is no actual object with those members, but rather a backend that will provide these values dynamically. Without these changes, I wouldn't be able to achieve this using DynamicExpresso, so I hope that it might be useful to others, too.

Let me know if you have any feedback on this.

### DIFF
--- a/src/DynamicExpresso.Core/DynamicExpresso.Core.csproj
+++ b/src/DynamicExpresso.Core/DynamicExpresso.Core.csproj
@@ -59,6 +59,8 @@
     <Compile Include="IdentifiersInfo.cs" />
     <Compile Include="LanguageConstants.cs" />
     <Compile Include="ParserArguments.cs" />
+    <Compile Include="Parsing\ResolveExpressionType.cs" />
+    <Compile Include="Parsing\ResolveExpressionEventArgs.cs" />
     <Compile Include="Parsing\Token.cs" />
     <Compile Include="Parsing\TokenId.cs" />
     <Compile Include="ReferenceType.cs" />
--- a/src/DynamicExpresso.Core/Interpreter.cs
+++ b/src/DynamicExpresso.Core/Interpreter.cs
@@ -442,8 +442,6 @@ namespace DynamicExpresso
 				throw new Exception("Detected identifiers doesn't match actual identifiers");
 			if (info.Types.Count() != lambda.Types.Count())
 				throw new Exception("Detected types doesn't match actual types");
-			if (info.UnknownIdentifiers.Count() != lambda.UsedParameters.Count())
-				throw new Exception("Detected unknown identifiers doesn't match actual parameters");
 		}
 #endif
 		#endregion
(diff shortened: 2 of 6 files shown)


davideicardi:
Thank you Adam. Nice feature. I will think about it when I will release the next version.

davide

davideicardi:
I'm studying your new feature...

Instead of having a generic ResolveUnknownExpression event, maybe can be better to provide some sort of [strategy](http://www.oodesign.com/strategy-pattern.html) with specific method for each case (property/field, method, ...).

Something like:

	public interface IResolveUnknownExpressionStrategy
	{
		Expression ResolveMethod(...);
		Expression ResolveProperty(...);
	}
	....

In this way we can have less parameters and a more extensible design. In the same way we can also support [dynamics](https://github.com/davideicardi/DynamicExpresso/issues/16) I think, with a custom implementation that implement dynamic (like a DynamicResolveStrategy class).

What do you think?

thanks

AdamsLair:
> What do you think?

That does sound like a better solution! It would probably be the easiest to reject my request and re-implement this, as I went down a different path here.

It might deal with dynamics as well, but the parser needs to be a little more flexible for that I think. Resolving methods and properties is not going to be enough for them - they can also be invoked like methods themselfs (could be delegates), indexed with any amount of parameters, be passed as arguments to other methods and indexers, and even allow any kind of operator.

davideicardi:
Yes, you are right, a full support of dynamics can be more complex.

Anyway for now I will reject this request, if you have time to work on the strategy solution I will be happy to merge it. Otherwise I can try to implement it in the future.

thanks again!

PR #33, opened 2015-07-17
Fixed performance issue where Interpreter's ctor ran too slow

OmerRaviv:
Hi,

I've been using DynamicExpresso indirectly via @Alex141 's CalcBinding library, and it is fantastic!
However, when using the library extensively I've found there to be a performance issue inside of Interpreter constructor. 

Here is a profiling snapshot from dotTrace:
image

I later created a test and found that creating 1000 Interpreter instances on my modern machine took just over 4 seconds.
 

I've eliminated the problem by adding a few fixes:
1. Switched to using a HashSet instead of a List
2. Only load the extension methods once per primitive type
3. Only attempt to read extension methods from System.Type's that represent static classes (both .IsSealed and .IsAbstract are true)

Added a unit test to prevent regression.

### DIFF
--- a/src/DynamicExpresso.Core/Interpreter.cs
+++ b/src/DynamicExpresso.Core/Interpreter.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 
 namespace DynamicExpresso
 {
@@ -278,13 +279,9 @@ namespace DynamicExpresso
 
 			_settings.KnownTypes[type.Name] = type;
 
-			var extensions = ReflectionExtensions.GetExtensionMethods(type.Type);
-			foreach (var extensionMethod in extensions)
+		    foreach (var extensionMethod in type.ExtensionMethods)
 			{
-				if (!_settings.ExtensionMethods.Contains(extensionMethod))
-				{
-					_settings.ExtensionMethods.Add(extensionMethod);
-				}
+			    _settings.ExtensionMethods.Add(extensionMethod);	
 			}
 
 			return this;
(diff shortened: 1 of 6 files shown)


Alex141:
Great perfomance fix! The only thing that I think may be improved - unit test, that depends from running time on concrete machine. I think if it possible, right way is to use mocks, that will count the number of calls. If it is impossible, nothing wrong)

@OmerRaviv can you write some xaml example of using CalcBinding, in which these delays have occurred?

OmerRaviv:
@Alex141 Thank you, and thanks for creating such a great library :)

[Here is a XAML sample WPF Application that reproduces the problem](https://gist.github.com/OmerRaviv/0dd8db462c1299f9cc2e). If you copy paste this into a blank new WPF application and hit F5, it will take a few seconds for the app to come up - if you comment out the line that uses CalcBinding, it will come up instantly.

As for the unit test, as per [this stackoverflow thread](http://stackoverflow.com/questions/15181358/how-can-i-unit-test-performance-optimisations-in-c)  my opinion is that using mocks for testing performance regression won't get us the desired effect - it will only be able to test that the particular underlying problem hasn't re-occurred, while I would prefer to define the test in terms of "this operation should take no more than ___" (some reasonable amount of time). I choose 200 millisec as a reasonable upper bound which should be OK for any reasonable machine.

While it's true that the test would fail on extremely slow machines or machines that are doing a lot of other work simultaneously,  and some would argue that a unit test that may randomly pass or fail based on external factors is not a proper unit test, I think a simple performance regression that is run repeatedly on the same CI configuration can get you a lot of value.

davideicardi:
Thanks @OmerRaviv  for your pull request. I will check it in the following days and merge it.

My original idea was to reuse the same Interpreter class when performance are important. In you case it is possible to reuse the same instance?

But in any case if the constructor can be optimized is a good improvement!

thanks

davideicardi:
Thank you @OmerRaviv , I have just merged your code and published package 1.3.1!

Sorry for the late response...

PR #46, opened 2016-08-25
Fix: >1 overload params[] methods suit given args

repko-artem:
I.e. expression is "Utils.Max(1, 2, 3)" and there are methods
`int Max(params int[] values)`
`long Max(params long[] values)`
In such case lib method Parser.FindBestMethod has more than 1 applicable methods and invokes method MethodHasPriority.
In this method there's no check if parameters have ParamsArrayType and it throws exception on method.Parameters[i] "Out of range" exception.
I've handled this situation in the method MethodHasPriority.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1622,6 +1622,13 @@ namespace DynamicExpresso.Parsing
 			return parameterInfo.IsDefined(typeof(ParamArrayAttribute), false);
 		}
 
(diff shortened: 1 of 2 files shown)


davideicardi:
Sorry for the late response, and thank you very much for your code. Seems fine for me.
Can I just ask you to add the corresponding unit test? Also an easy one...

repko-artem:
I've written the test. Hope it's ok.

davideicardi:
I just merged your changes and released a new version 1.3.2.

Thanks!

repko-artem:
Thank you, Davide!

Issue #43, opened 2016-06-20
Unable to evaluate expression if a number starts with a `.` (dot)


sgokuladas:
The interpreter cannot evaluate a formula if I say interpreter.evaluate("1 + .2")

Note:  interpreter.evaluate("1 + **0**.2") works

sgokuladas:
I'm new around here, but I'll give it a go if I get some time next week




On Mon, Jun 20, 2016 at 11:57 PM +1000, "Davide Icardi" <notifications@github.com<mailto:notifications@github.com>> wrote:


You are right, this is a bug. I added this is my todo list but I don't know when I can work on it.
Eventually I appreciate any help (pull request).

-

JustinRChou:
Solution might not be as clean without changing more code.

davideicardi:
Fixed, thanks to @JustinRChou

PR #50, opened 2016-10-08
Fix for #43 when .2 was not parsing is a Real Literal.

JustinRChou:
Fix for #43 when .2 was not parsing is a Real Literal.
Added additional Unit Tests for the new parsing.
Not sure what went wrong is the spacings.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -16,2113 +16,2089 @@ using System.Text;
 
 namespace DynamicExpresso.Parsing
 {
-    internal class Parser
-    {
-        public static Expression Parse(ParserArguments arguments)
-        {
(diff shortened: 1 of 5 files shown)


davideicardi:
Thanks very much for your pull request! 
Can I just ask you do leave the tabs instead of changing it to spaces? Otherwise we have hundreds of these kind of changes. I usually prefer tabs because are more configurable for each developer, but I know that this is a subjective argument 😄 

I think that there are some configuration on git that can handle these scenarios...

JustinRChou:
Seems like I fixed the spacing diffs.

And it turns out that my fix was failing on the new unit tests for .2f and .2m formats.

davideicardi:
Thanks @JustinRChou !
I will integrate these changes in the next release.

davideicardi:
Thanks, merged in version 1.3.4.7

Issue #36, opened 2015-10-29
Localizable Error Messages


Raphael-Saldanha:
It would be nice if it was possible to replace error messages. My system is used in several languages and i need to show all error messages in user's language.

The simplest solution is to change ErrorMessages from internal to public and remove const modifier from error messages.

But it would be nice if we have a mechanism to inject error messages or a Abstract Factory that creates ErrorMessages based on some conditions.

davideicardi:
Yes, it can be a future enhancement.

For the solution I think that the most easy way is to use localized .NET resource files: [How to: Create a Localized Version of a Resource File](https://msdn.microsoft.com/en-us/library/vstudio/aa992030(v=vs.100).aspx). So instead of having a ErrorMessages class we should have a ErrorMessages resource file.

Of course any help is appreciated :wink:

SaldanhaBR:
I will provide a possible solution soon.

davideicardi:
Fixed, thanks to @jasonhumberstone

PR #51, opened 2016-10-15
Moved parser error messages into a resource file for issue #36

jasonhumberstone:
This is a change to facilitate the localisation of parser error messages for issue #36. New resource files for each supported language and region can now be created and an appropriate satellite DLL will be created for each file upon compilation. 

For example, a resource file for the French language in Canada would be called 'ErrorMessages.fr-CA.resx', and the system will try to load the corresponding DLL at run time on systems with that locale setting. If the appropriate DLL isn't found, it will fallback to the default resource file (called 'ErrorMessages.resx') I created in my commit. 

PS: Apologies for the duplicate commit reference on the issue. I made a mistake as this is my first time on GitHub.


davideicardi:
Thank you @jasonhumberstone ! Good idea!

I will integrate these changes in the next release.

davideicardi:
Thanks, merged in version 1.3.4.7

PR #74, opened 2018-03-08
Variables and parameters should take precedence over known types.

mjamin:
Ran into this issue while trying to use `Object` as a variable name.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -623,12 +623,6 @@ namespace DynamicExpresso.Parsing
 			if (_token.text == ParserConstants.KEYWORD_TYPEOF)
 				return ParseTypeof();
 
-			Type knownType;
-			if (_arguments.TryGetKnownType(_token.text, out knownType))
-			{
-				return ParseTypeKeyword(knownType);
-			}
-
 			Expression keywordExpression;
 			if (_arguments.TryGetIdentifier(_token.text, out keywordExpression))
 			{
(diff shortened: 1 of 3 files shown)


davideicardi:
Thanks very much for the pull request. I just not sure if this is correct. In C# I think that you cannot name a variable like this, and I usually try to be compatible with C# when possible.

Eventually this feature can be turned on with a configuration?

mjamin:
Makes sense from that viewpoint.

But identifiers generally can be named like types and that's where I ran into this issue. I'm evaluating a script in a context that's defined by an instance of an object. I make all fields and properties of that object available via `SetVariable` and all methods via `SetFunction`. In my case, that context object happens to have a property named `Object`.

I think these are the options here:

- Do nothing as everything technically works as intended.
- Rename `SetVariable` to `SetIdentifier` and change the precedence.
- Add a new `SetContext` method that takes an object and exposes its public and/or private members to the interpreter.

davideicardi:
Maybe we can add some extension methods, one to simplify the usage of `SetIdentifier` and one to simulate a `SetContext`. I think that all features can be added without changing the core `Interpreter` class. Right?
I try to leave the `Interpreter` class as light as possible to allow easy refactoring.

Like:

	public static class InterpreterExtensions
	{
		public static Interpreter SetIdentifier<T>(this Interpreter interpreter, string name, T value)
		{
			return interpreter.SetIdentifier(
				new Identifier(name, Expression.Constant(value)));
		}

		public static Interpreter SetContext(this Interpreter interpreter, object value)
		{
			// TODO
		}
	}

mjamin:
Sounds good. I'll close this and submit another PR when I get around to it.

PR #80, opened 2018-07-03
Support calling functions with default parameters

weltall:
At the moment, calling a method with default arguments without defining them explicitly causes a failure:

Consider this example:

```
    class Program
    {
        class test
        {
            public string abc(int a, int b = 5, int c = 10)
            {
                return (a + b +c).ToString();
            }
        }
        static void Main(string[] args)
        {
            Interpreter a = new Interpreter();
            a.SetVariable("Test", new test());
            Console.WriteLine(a.Eval<string>("Test.abc(2)")); //fails
            Console.WriteLine(a.Eval<string>("Test.abc(2,5)")); //fails
            Console.WriteLine(a.Eval<string>("Test.abc(2,4,11)")); //ok
            //Console.WriteLine(a.Eval<string>("Test.abc(2,4,11,12)")); // correct to be fail
            //Console.WriteLine(a.Eval<string>("Test.abc()")); // correct to be fail
            Console.ReadKey();
        }
    }
```

The first and second eval will fail with an exception saying no suitable method was found in Test, while clearly abc(int, int = 5, int = 10) is suitable for both. 

The patch allows these two cases to be handled by filling the missing parameters with their default values, if available, while preparing the method.

The third eval worked, and still works with the patch, as expected.

The forth eval didn't work, and still doesn't work with the patch, as expected.

The fifth eval didn't work, and still doesn't work with the patch, as expected as the first parameter doesn't have a default value defined.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1280,7 +1280,7 @@ namespace DynamicExpresso.Parsing
 
 		private static bool CheckIfMethodIsApplicableAndPrepareIt(MethodData method, Expression[] args)
 		{
-			if (method.Parameters.Length > args.Length)
+			if (method.Parameters.Count(y => !y.HasDefaultValue) > args.Length)
 				return false;
 
 			var promotedArgs = new List<Expression>();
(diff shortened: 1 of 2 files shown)


davideicardi:
Thanks! Very nice enhancement! I will merge it soon.

All other unit tests are passing? 
If you can provide some unit tests I can merge it directly...otherwise I will write it whenever possible.

thanks!

weltall:
Yes they seems to be all passing fine.
I've added some unit test for the methods with optional parameter.

davideicardi:
Just merged and released 2.1.0. Thanks! 🎉

PR #83, opened 2018-07-25
Adding support to inject custom converters

iarly:
My expression has integer numbers operating with nullable decimals. In this case, an expression compatibility problem happens when I parse a simple multiplication: x * 100, where x is a nullable decimal parameter.

Then, I needed to extend the conversion behavior to enable custom number conversions that can be passed by the developer through parameters. 

After this implementation, I solved my problem writing a custom IntegerConverter in my project. This custom IntegerConverter converts all numbers to decimal.

### DIFF
--- a/src/DynamicExpresso.Core/Converters/IConverter.cs
+++ b/src/DynamicExpresso.Core/Converters/IConverter.cs
@@ -1,0 +1,7 @@
+﻿namespace DynamicExpresso.Converters
+{
+	public interface IConverter
+	{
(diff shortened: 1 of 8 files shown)


nike61:
@Iarly check my PR: https://github.com/davideicardi/DynamicExpresso/pull/89. Maybe it will help you

davideicardi:
I will close the PR for inactivity. Feel free to reopen if you still need this PR.

PR #87, opened 2018-09-20
Added possibility to use new Resources Files

jbravobr:
I've added the option to use new resource files. As my project deals with the Brazilian Portuguese language, we need to extend the library so that it is possible to configure which language the exceptions will be displayed, if there is a resource for it.

We do it by implement a Helper and did some overloads onto the main methods from the Parser. We also added a new resource, with all the translations needed for Brazilian Portuguese.

### DIFF
--- a/src/DynamicExpresso.Core/DynamicExpresso.Core.csproj
+++ b/src/DynamicExpresso.Core/DynamicExpresso.Core.csproj
@@ -24,4 +24,19 @@
 		<Reference Include="Microsoft.CSharp" />
 	</ItemGroup>
 
+	<ItemGroup>
+	  <Compile Update="Resources\ErrorMessages.Designer.cs">
+	    <DesignTime>True</DesignTime>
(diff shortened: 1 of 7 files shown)


davideicardi:
Thanks very much @jbravobr, @gztavo . 

It also works with .NET Core?
Can you add an unit test for each language to try it?

Finally can I ask you to leave the tab/space formatting as is to reduce differences and better understand your changes.

thanks!

jbravobr:
Oi @davideicardi!

Obrigado pela revisão tão cedo! Sim, funciona com o .NET Core, na verdade estamos usando uma API WEB do núcleo .NET e com o Projeto padrão do núcleo .NET.

Vou adicionar alguns testes de unidade para os idiomas que existem no momento

davideicardi:
Closing for inactivity for now. Feel free to reopen in case you are able to fix formatting.

PR #89, opened 2018-09-26
FIX #88 calculations not working with non nullables and nullables numbers

### LINKED ISSUE #88: a + b when a is of type Nullable<double> and b is of type Int32
Can't find solution to make this work properly:
```
interpreter.SetVariable("a", 1, typeof(int));
interpreter.SetVariable("b", 1.2, typeof(double?));
var q = interpreter.Eval("a + b");
```
I'm getting
> System.InvalidOperationException: "The binary operator Add is not defined for the types 'System.Int32' and 'System.Nullable`1[System.Double]'."

nike61:
FIX #88

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1423,7 +1423,7 @@ namespace DynamicExpresso.Parsing
 				}
 			}
 
-			if (type.IsGenericType)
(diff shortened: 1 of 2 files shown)


davideicardi:
Thanks @nike61 ! I will check it.

NielsUll:
I too would appreciate this getting released

davideicardi:
Thanks @nike61, just release v2.2.0 with this fix!

PR #105, opened 2019-02-02
Handle default parameters type explicitly

weltall:
When an argument default value is gathered from a method definition be sure to pass also its associated argument when creating the Expression.Constant.

This avoids a problem which happens when the argument is null: in this case the type of the argument will be System.Object and that will fail when binding the arguments to a method defining a different type (for example string)

Additionally see remarks section from https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.constant?view=netframework-4.7.2#System_Linq_Expressions_Expression_Constant_System_Object_ )

In addition a new unit test was added to showcase this situation.


davideicardi:
Thanks @weltall , just merged!

PR #106, opened 2019-02-03
Parser.cs splitting

btnlq:
`Parser.cs` contains 2237 lines, so I tried to split it into three files/classes:

- `Parser.cs` (1307 lines) - methods that use and change the state of the parser
- `Expressions.cs` (663 lines) - methods that create, change and use expressions
- `Types.cs` (287 lines) - methods that get information about types

What do you think about this?


davideicardi:
Thanks @btnlq . Yes, you are right, `Parser.cs` is too big.

Just not sure if this is the right splitting. Or maybe we can just rename `Expressions.cs` and `Types.cs` with something more related to the project scope. Let me think some days...

davideicardi:
Closing for now, feel free to reopen in case you think it is still relevant.

Issue #86, opened 2018-09-19
Support verbatim strings (@"Hello, world!")


jubba2001:
Please, support C# verbatim strings:

Assert.AreEqual(@"Hello, world!", target.Eval(@"@""Hello, world!"""));

davideicardi:
Thanks, @jubba2001  I will think about this. Pull requests are appreciated :grinning:

PR #109, opened 2019-05-13
Support verbatim strings (@"Hello, world!") #86

pranav700:


pranav700:
Yes, I think escape chars are not supported but will add couple of unit tests around that in my next PR.
On point 1, what do you mean by same tab config here?
Thanks

davideicardi:
@pranav700 
Sorry for the late response...

> On point 1, what do you mean by same tab config here?
I mean that I have used `tab` instead of `spaces` for indenting code in this project. If you can I prefer if you align with that, to avoid mixed indentation...

Also can you please explain me your use case? Why do you need vermatim strings? Just to understand better your requirements...

davideicardi:
Closing for inactivity for now. Feel freet to reopen it in case it is still relevant.

PR #114, opened 2019-09-19
added support for null enum

igor-smi:
When parsing an expression containing a type of null enum, an error occurs. I tried to fix it.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/ParseSignatures.cs
+++ b/src/DynamicExpresso.Core/Parsing/ParseSignatures.cs
@@ -79,6 +79,11 @@ namespace DynamicExpresso.Parsing
 			void F(bool? x);
 		}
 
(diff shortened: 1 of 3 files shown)


davideicardi:
Thanks @igor-smi !
Can I ask you to include a unit test to test and describe the new feature?

igor-smi:
Added tests.

In what format and where should the function description be included?

davideicardi:
@igor-smi 
Sorry for late response...

> In what format and where should the function description be included?
I mean if you can just explain here in the PR a little better your problem and solution. Just to understand exactly your use case.

Also can you please use tab indentation like the rest of the project? Just to avoid mixed indentation...
thanks!

davideicardi:
I think that this problem is now fixed with #124 . Open again the PR in case it is not yet fixed...

PR #117, opened 2020-01-22
Allow assignment using implicit operator cast

mtone:
In ParseAssignment, instead of throwing immediately when PromoteExpression returns null, calls ConvertChecked -- which I got from https://github.com/davideicardi/DynamicExpresso/issues/7

This enables support for implicit operator casts in assignments. I wouldn't expect this to break anything, given it would otherwise throw.

Use case: using this library in a partial trust environment (MS Dynamics 365) where reflection support is limited and the dynamic type is not allowed. As an alternative, I'm using a boxed value type with ImplicitOperator to handle specific type conversions/comparisons/assignments I want to support (to store business logic configuration). In .NET standard mode, everything works nicely.

Thanks for the great library!

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -102,9 +102,19 @@ namespace DynamicExpresso.Parsing
 
 				var right = ParseAssignment();
 				var promoted = PromoteExpression(right, left.Type, true);
-				if (promoted == null)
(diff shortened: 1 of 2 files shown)


davideicardi:
Thanks @mtone !
Sorry for late response.

Can you just also add a test to verify what happen if the `ConvertChecked` is not possible? Like assignin it to a type that it is not listed as implicit. Just to better understand any possible drawback...

davideicardi:
Closing for inactivity for now. Feel free to reopen if it is still relevant.

PR #134, opened 2021-02-22
Dynamic Expresson fix parsing of Exponential format number

### LINKED ISSUE #133: ParseLiteralDecimalNumberStyle
I had errors parsing scientific formatted numbers. How about adding NumberStyles.AllowExponent to Parser?

private const NumberStyles ParseLiteralDecimalNumberStyle = NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent;

alexk47:
Fixes #133

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -27,6 +27,7 @@ namespace DynamicExpresso.Parsing
 		private const NumberStyles ParseLiteralNumberStyle = NumberStyles.AllowLeadingSign;
 		private const NumberStyles ParseLiteralUnsignedNumberStyle = NumberStyles.AllowLeadingSign;
 		private const NumberStyles ParseLiteralDecimalNumberStyle = NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint;
+        private const NumberStyles ParseLiteralDoubleNumberStyle = NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent;
 		private static readonly CultureInfo ParseCulture = CultureInfo.InvariantCulture;
 
 		private readonly ParserArguments _arguments;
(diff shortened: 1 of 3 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -27,6 +27,7 @@ public static Expression Parse(ParserArguments arguments)
 		private const NumberStyles ParseLiteralNumberStyle = NumberStyles.AllowLeadingSign;
 		private const NumberStyles ParseLiteralUnsignedNumberStyle = NumberStyles.AllowLeadingSign;
 		private const NumberStyles ParseLiteralDecimalNumberStyle = NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint;
+        private const NumberStyles ParseLiteralDoubleNumberStyle = NumberStyles.AllowLeadingSign | NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent;
```
davideicardi, line 30: Can you just align the code using the same indentation?

davideicardi (changes requested):
Thanks very much! Just a couple of notes:

- can you add a some more tests where we use this notation inside more complex expressions? I have the feealing that some can not work ... (like having a math expression with `+` or `-` ...)
- do you think that it makes sense as a default or we should put a configuration instead?

src/DynamicExpresso.Core/Parsing/Parser.cs
alexk47: code aligned

alexk47: >I have the feealing that in some cases we can have problems ... (like having a math expression with + or - ...)

in my project  i use expression like that
@Var1 + @Var2 / Math.Pow(2, 2)
where @Var1 = 2e+10 and @Var2 = 0.0006
with fix which i've made, all my expressions calculated fine

>do you think that it makes sense as a default or we should put a configuration instead?

i don't think so, because exponential format is one default formats of number reprecentation, you can write 0.0003 or 3e-4  and this are same numbers

alexk47: >can you add a some more tests where we use this notation inside more complex expressions?

where i can add new tests, which class i should extends. I can add new test to GithubIssues.cs

davideicardi: Very good! Thank you! In the week-end I will do a final check and release a new version.

davideicardi:
v2.4.0 released! Thank you for the help!

alexk47:
> v2.4.0 released! Thank you for the help!
You are welcome

Issue #135, opened 2021-04-22
Custom operators not resolved when called on derived instances


metoule:
To illustrate the issue, let's imagine I have the following hierarchy:

```c#
public class Base 
{
  public double Value { get; set; }
  public static Base operator +(Base left, Base right)
  {
    return new Base { Value = left.Value + right.Value };
  }
}

public class Derived : Base
{
}
```

With the proper C# compiler, it's possible to add two `Derived` instance, because the operator `Base.operator+` exists.

However, if I try to perform this operation with DynamicExpressso, I get a `ParseException`:

> The binary operator Add is not defined for the types 'Derived' and 'Derived'.

Test case:

```C#
static void Main()
{
  var left = new Derived { Value = 2 };
  var right = new Derived { Value = 5 };
  
  // properly resolved by the C# compiler
  var res1 = left + right ;

  var interpreter = new Interpreter();
  interpreter.SetVariable("left", left);
  interpreter.SetVariable("right", right);

  // --> this fails
  var parsedExpression = interpreter.Parse("left + right");
  var result = parsedExpression.Invoke() as Base;
}
```

Note that the operator resolution works if at least one of the operand has the `Base` type.

davideicardi:
Thank you for the bug report @metoule . It is probably some bug related to how the operators are resolved.
I will mark it as a bug, any help is approciated.

metoule:
I'm looking at it, and it seems like the bug it in the way the operators are resolved in the Linq expressions. 
For example, if I strip the issue down to the Linq expressions: 

````
var m1 = new Derived { Value = 1 };
var m2 = new Derived { Value = 2 };

var left = Expression.Constant(m1);
var right = Expression.Constant(m2);
var expr = Expression.Add(left, right);
````

the bug occurs, because Linq doesn't check the base classes for the operators.

I can solve the issue by specifying the method to use:

````
var method = typeof(Base).GetMethod("op_Addition");
var expr = Expression.Add(left, right, method);
````

I'm not sure how easy it is to fix this on DynamicExpresso side, so I'll head over to `dotnet/runtime` to create an issue.

davideicardi:
Fixed by #136  . Thanks!

PR #136, opened 2021-04-22
Unary and binary operators: find the proper overload by also checking the base classes

metoule:
This is an attempt at fixing #135 . The gist is to try to find the proper user defined operator instead of solely relying on the Linq resolution (which doesn't check the base classes).

There were already some tests for user defined operators, I've added a use case to expand on it.


davideicardi (approved):
Great work! Thank you!

davideicardi:
Released v2.4.1

PR #137, opened 2021-05-10
Fix user defined operator resolution

metoule:
This is a follow up of my previous PR #135 . If a user defined operator was only defined on the right operand's type, it was not taken into account..


davideicardi (approved):
LGTM! Thanks again!

PR #138, opened 2021-05-10
Allow the registration of function overloads

### LINKED ISSUE #82: Allow to register function with the same name but different arguments
Allow to register 2 functions with the same name but with a different number of arguments:

````
Func<decimal, decimal> roundFunction1 = (userNumber) => Math.Round(userNumber);
Func<decimal, int, decimal> roundFunction2 = (userNumber, decimals) => Math.Round(userNumber, decimals);
interpreter.SetFunction("ROUND", roundFunction1);
interpreter.SetFunction("ROUND", roundFunction2);
````

Currently only the last `SetFunction` call will be preserved.

See also #81 .

### PR DESCRIPTION
**Pull request by metoule**
This is an attempt to fix #82, without changing the public interface. 

The idea is to use a custom identifier for functions, that rely on a dummy, custom, Linq Expression that represents the list of possible delegates. During the method invocation resolution, we detect this custom Expression, and check each possible overload to see if it matches the argument list.

The unit test is:

```
[Test]
public void Keywords_with_overloaded_delegates()
{
	Func<decimal, decimal> roundFunction1 = (userNumber) => Math.Round(userNumber);
	Func<decimal, int, decimal> roundFunction2 = (userNumber, decimals) => Math.Round(userNumber, decimals);

	interpreter.SetFunction("ROUND", roundFunction1);
	interpreter.SetFunction("ROUND", roundFunction2);

	Assert.AreEqual(3.13M, interpreter.Eval("ROUND(3.12789M, 2)"));
	Assert.AreEqual(3M, interpreter.Eval("ROUND(3.12789M)"));
}
```

### DIFF
--- a/src/DynamicExpresso.Core/Identifier.cs
+++ b/src/DynamicExpresso.Core/Identifier.cs
@@ -39,9 +39,15 @@ namespace DynamicExpresso
 	/// </summary>
 	internal class MethodGroupExpression : Expression
 	{
-		private readonly List<Expression> _overloads = new List<Expression>();
+		private readonly List<Delegate> _overloads = new List<Delegate>();
 
-		internal IReadOnlyCollection<Expression> Overloads => _overloads.AsReadOnly();
+		internal IReadOnlyCollection<Delegate> Overloads
+		{
+			get
+			{
+				return _overloads.AsReadOnly();
+			}
+		}
 
 		internal MethodGroupExpression(Delegate overload)
 		{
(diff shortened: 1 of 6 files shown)


davideicardi (approved):
LGTM! Good Idea!

It should also be compatibility with old versions, right? I don't see any problem... If you confirm that I will merge it.

metoule:
I didn't change the public interface, so it should be compatible with the old versions. I can't see any use case that would impact current users.

Gonzo345:
Wow, this is huge! Thanks!

PR #145, opened 2021-05-24
Fix #144: consider a delegate's invoke methods

### LINKED ISSUE #144: Throw error when parsing the method with default parameters in v2.5.0
Hello, I have written a little test, it will pass in `v2.4.1` and failed in `v2.5.0`. But I can not find any breaking change in the release log, 
Is this a bug or am I missing something?  Thank you!

### Test case:
```csharp
private delegate bool GFunction(string arg = null);

[Fact]
public void TestParseGFunction()
{
    static bool GetGFunction(string arg = null)
    {
        return arg is not null;
    };

    interpreter.SetFunction("GFunction", (GFunction) GetGFunction);
    interpreter.SetVariable("arg", "arg");

    Assert.True((bool) interpreter.Eval("GFunction(arg)"));
    Assert.False((bool) interpreter.Eval("GFunction()"));
}
```

### Failed information:
```csharp
DynamicExpresso.Exceptions.ParseException
Argument list incompatible with delegate expression (at index 0).
   at DynamicExpresso.Parsing.Parser.ParseMethodGroupInvocation(MethodGroupExpression methodGroup, Int32 errorPos)
   at DynamicExpresso.Parsing.Parser.ParseUnary()
   at DynamicExpresso.Parsing.Parser.ParseMultiplicative()
   at DynamicExpresso.Parsing.Parser.ParseAdditive()
   at DynamicExpresso.Parsing.Parser.ParseTypeTesting()
   at DynamicExpresso.Parsing.Parser.ParseComparison()
   at DynamicExpresso.Parsing.Parser.ParseLogicalAnd()
   at DynamicExpresso.Parsing.Parser.ParseLogicalXor()
   at DynamicExpresso.Parsing.Parser.ParseLogicalOr()
   at DynamicExpresso.Parsing.Parser.ParseConditionalAnd()
   at DynamicExpresso.Parsing.Parser.ParseConditionalOr()
   at DynamicExpresso.Parsing.Parser.ParseConditional()
   at DynamicExpresso.Parsing.Parser.ParseAssignment()
   at DynamicExpresso.Parsing.Parser.ParseExpressionSegment()
   at DynamicExpresso.Parsing.Parser.ParseExpressionSegment(Type returnType)
   at DynamicExpresso.Parsing.Parser.Parse()
   at DynamicExpresso.Interpreter.ParseAsLambda(String expressionText, Type expressionType, Parameter[] parameters)
   at DynamicExpresso.Interpreter.Parse(String expressionText, Type expressionType, Parameter[] parameters)
   at DynamicExpresso.Interpreter.Eval(String expressionText, Type expressionType, Parameter[] parameters)
   at DynamicExpresso.Interpreter.Eval(String expressionText, Parameter[] parameters)
   at NetCasbin.UnitTest.UtilityTest.TestParseGFunction() in C:\Users\Sagilio\source\repos\OnWork\Casbin\Casbin.NET\NetCasbin.UnitTest\UtilityTest.cs:line 26
```

metoule:
This fixes #144 by taking a delegate's invoke methods into account while resolving the method invocation of a function defined via `SetFunction`.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Dynamic;
 using System.Globalization;
@@ -931,7 +931,22 @@ namespace DynamicExpresso.Parsing
 			var args = ParseArgumentList();
 
 			// find the best delegates that can be used with the provided arguments
-			var applicableMethods = FindBestMethod(methodGroup.Overloads.Select(_ => _.Method), args);
+			var flags = BindingFlags.Public | BindingFlags.Instance;
+			var candidates = methodGroup.Overloads
+				.Select(_ => new
+				{
+					Delegate = _,
+					Method = _.Method,
+					InvokeMethods = _.GetType().FindMembers(MemberTypes.Method, flags, Type.FilterName, "Invoke").Cast<MethodInfo>(),
+				})
+				.ToList();
+
+			var applicableMethods = FindBestMethod(candidates.SelectMany(_ => _.InvokeMethods), args);
+
+			// no method found: retry with the delegate's method
+			// (the parameters might be different, e.g. params array, default value, etc)
+			if (applicableMethods.Length == 0)
+				applicableMethods = FindBestMethod(candidates.Select(_ => _.Method), args);
 
 			if (applicableMethods.Length == 0)
 				throw CreateParseException(errorPos, ErrorMessages.ArgsIncompatibleWithDelegate);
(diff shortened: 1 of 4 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -940,8 +955,8 @@ private Expression ParseMethodGroupInvocation(MethodGroupExpression methodGroup,
 				throw CreateParseException(errorPos, ErrorMessages.AmbiguousDelegateInvocation);
 
 			var applicableMethod = applicableMethods[0];
-			var delegateExp = Expression.Constant(methodGroup.Overloads.Single(_ => _.Method == applicableMethod.MethodBase));
-			return Expression.Invoke(delegateExp, applicableMethod.PromotedParameters);
+			var usedDeledate = candidates.Single(_ => new[] { _.Method }.Concat(_.InvokeMethods).Any(m => m == applicableMethod.MethodBase)).Delegate;
```
davideicardi, line 958: Calling `Single` is safe? I mean there aren't cases where an exception can be throw? Maybe if there are more than 1 match?

davideicardi (commented):
LGTM!

src/DynamicExpresso.Core/Parsing/Parser.cs
metoule: I believe it's safe, because at this point there's a single `applicableMethod`, which comes from one of the delegate. There can be only one matching delegate, because otherwise there would be multiple `applicableMethod`.

davideicardi (approved):
Thanks!

Issue #149, opened 2021-06-07
Support for external ParameterExpression in Parameter constructor


nemo048:
Hello! 

We would like to have a an ability to set `ParameterExpression` in `Parameter` class. When this library is used with externally created Expressions, there is no way to pass a `ParameterExpression` into `Parameter` constructor.

Currently we have to use `ExpressionVisitor` instance to substitute the parameter generated in default constructor for our external parameter like this:
```C#
ParameterExpression externalParameter = Expression.Parameter( someType, "p" );

Parameter param = new Parameter(externalParameter.Name, externalParameter.Type);
ParameterExpression paremeterToReplace = param.Expression;

Lambda expr = new Interpreter().Parse("some expression", param);

Expression finalExpression = new ReplaceExpressionVisitor(paremeterToReplace, externalParameter).Visit(expr.Expression);
```

This could be solved by creating additional constructor that accepts `ParameterExpression` as an argument, so that the above code would look like this:

```C#
Parameter param = new Parameter(externalParameter);
Expression finalExpression = new Interpreter().Parse("some expression", param).Expression;
```

davideicardi:
Thanks! Can you please just explain what it is high level use case for this?
I mean why you should need an external expression?

Rebel028:
This external parameter may be a part of another previously created expression, that exists in the same scope. That previously expression may or may not be created via `Interpreter`, all we know that it uses the same parameter expression. 

It may be helpful if the lambda we create using `Interpreter` is a part of a bigger and complex expression.

davideicardi:
Closed by #150

PR #150, opened 2021-06-07
Added new constructor in Parameter with ParameterExpression

nemo048:
Solves the issue #149


Rebel028:
@davideicardi could you take a look at this one too? (Not sure if you got the notification and the PR is not linked to an actual Issue despite it's being mentioned in the comment here). 

Thanks!

davideicardi (approved):
Thank you for the improvement!

PR #152, opened 2021-06-07
Refactored method GenerateStringConcat in Parser.cs

### LINKED ISSUE #151: Concatenation of two strings is failing when used with latest EntityFramework Core
We've faced an issue while using `Interpreter` with a latest (5.0.6) EF Core. 

`Parser` is using `String.Concat(Object, Object)` method to concatenate two objects as strings. Unfortunately it could not be translated to SQL by EF Core and since version 3.1 it will throw an exception as Linq queries are no longer evaluated on client.

Instead of using  `String.Concat(Object, Object)` it would be better to use `String.Concat(String, String)` explicitly. 

First of all, it will make things made in #58  work, and additionally it will be able to use generated expression with Linq to Entities as EF Core can correctly translate `String.Concat(String, String)` to SQL

nemo048:
Fixes #151 

It would be great if you could review my changes for Parser.GenerateStringConcat method in this pull request.

PS. I also added two static fields for MethodInfos you use in there so that they won't have to be calculated on each call

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1346,7 +1346,7 @@ namespace DynamicExpresso.Parsing
 
 		private static bool IsNullableType(Type type)
 		{
-			return type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
+			return Nullable.GetUnderlyingType(type) != null;
 		}
 
 		private static bool IsDynamicType(Type type)
@@ -2126,25 +2126,25 @@ namespace DynamicExpresso.Parsing
 
 		private static Expression ToStringOrNull(Expression expression)
 		{
-			var nullConstant = Expression.Constant(null);
+			var nullableExpression = IsNullableType(expression.Type) ?
+				expression :
+				GenerateNullableTypeConversion(expression);
+
+			var expressionTypeNullConstant = Expression.Constant(null, nullableExpression.Type);
+			var stringNullConstant = Expression.Constant(null, typeof(string));
+
 			Expression condition = Expression.Equal(
-				expression,
-				nullConstant);
+				nullableExpression,
+				expressionTypeNullConstant);
 
 			return Expression.Condition(
 				condition,
-				Expression.ConvertChecked(nullConstant, typeof(String)),
+				stringNullConstant,
 				Expression.Call(expression, _toStringMethod));
 		}
 
 		private static Expression GenerateStringConcatOperand(Expression expression)
 		{
-			if (expression is ConstantExpression constantExpression &&
-			    constantExpression.Value == null)
-			{
-				return Expression.ConvertChecked(constantExpression, typeof(string));
-			}
-
 			return expression.Type != typeof(string)
 				? ToStringOrNull(expression)
 				: expression;
@@ -2536,12 +2536,13 @@ namespace DynamicExpresso.Parsing
 		{
 			var exprType = expr.Type;
 
-			if (exprType.IsGenericType && exprType.GetGenericTypeDefinition() == typeof(Nullable<>))
+			if (IsNullableType(exprType) ||
+			    !exprType.IsValueType)
 			{
 				return expr;
 			}
 
-			var conversionType = typeof(Nullable<>).MakeGenericType(expr.Type);
+			var conversionType = typeof(Nullable<>).MakeGenericType(exprType);
 			return Expression.ConvertChecked(expr, conversionType);
 		}
 
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -246,8 +246,8 @@ namespace DynamicExpresso.Parsing
 		{
 			var left = ParseTypeTesting();
 			while (_token.id == TokenId.DoubleEqual || _token.id == TokenId.ExclamationEqual ||
-					_token.id == TokenId.GreaterThan || _token.id == TokenId.GreaterThanEqual ||
-					_token.id == TokenId.LessThan || _token.id == TokenId.LessThanEqual)
+			       _token.id == TokenId.GreaterThan || _token.id == TokenId.GreaterThanEqual ||
+			       _token.id == TokenId.LessThan || _token.id == TokenId.LessThanEqual)
 			{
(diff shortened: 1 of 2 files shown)


davideicardi:
Great! Thank you!

Can you just provide an unit test to verify that? Maybe concateneting a string with some other object? Or do some similar test?
Potentially you can also test the generated expression. What do you think?

nemo048:
Should I add a reference to Entity Framework Core to unit tests project or I just need to test the existing method?

Also there is already a couple of tests with concatenation of two strings and a string with an integer. Do you need anything more than this?

davideicardi:
No, I don't think that we have to reference Entity Framework. But in some way we should test the expected behaviour.

Maybe you can just verify that calling `Concat`  will result in calling `String.Concat(String, String)` instead of `String.Concat(Object, Object)`? I think that you can do that by inspecting the resulted expression. Right?

Also what happen if we have a `null`? It is supported?

nemo048:
I've added two more tests: one is asserting that we call `String.Concat(String, String)` method and another one is checking null to string concatenation behaviour. And yes, now null is supported 😄

nemo048:
Sorry 😄  Could you please say is it enough to merge or should I do something more?

test/DynamicExpresso.UnitTest/OperatorsTest.cs
```diff
@@ -130,6 +133,34 @@ public void String_Concatenation_with_type_conversion()
...
+        [Test]
+        public void String_Concatenation_with_null()
+        {
+            Interpreter interpreter = new Interpreter();
+
+            string expressionText = "\"ciao \" + null";
+
+            Assert.AreEqual("ciao ", interpreter.Eval(expressionText));
+        }
+
```
davideicardi, line 163: What if you write:
```
        [Test]
        public void String_Concatenation_with_null()
        {
            Interpreter interpreter = new Interpreter();
            string expressionTextWithNullKeyword = "\"ciao \" + null";
            Assert.AreEqual("ciao ", interpreter.Eval(expressionTextWithNullKeyword));

            // create a function that returns null
            Func<String> someFunc = () => null;
            interpreter.SetFunction("someFunc", someFunc);
            string expressionTextWithNullFunc = "\"ciao \" + someFunc()";
            Assert.AreEqual("ciao ", interpreter.Eval(expressionTextWithNullFunc));
        }
```
My concern is that we should not call `.ToString` directly, because on null it will not work. But checking if the expression is null I think is not enough.
Maybe there is some better solution, but I need to think about it.

Also, one minor fix, sorry ;-) : can you please align code space style?
The repo is configured using tabs, I think that you have used spaces.

davideicardi (changes requested):
Some considerations about null expression, and one small indentation problem.
Thanks!

metoule:
It might also be a good idea to add a test with a variable of a type that overrides the `ToString` method, to ensure the overridden method is called (and not the base implementation). Something like this:

```csharp
public class MyClass
{
	public override string ToString()
	{
		return "MyClassStr";
	}
}

[Test]
public void String_Concatenation_with_overridden_ToString()
{
	Interpreter interpreter = new Interpreter()
		.SetVariable("myClass", new MyClass());
		
	Assert.AreEqual("ciao MyClassStr", interpreter.Eval("\"ciao \" + myClass"));
}
```

nemo048:
Hello @davideicardi and @metoule 😄 Sorry for my late response. I fixed indentation (hopefully), added more tests for concatenation with null and also added concatenation tests for classes with overridden ToString() method. Could you please review my changes?

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -2090,23 +2115,39 @@ private MethodData FindBinaryOperator(string operatorName, Expression left, Expr
...
+
+			return Expression.Condition(
+				condition,
+				Expression.ConvertChecked(nullConstant, typeof(String)),
+				Expression.Call(expression, _toStringMethod));
+		}
+
+		private static Expression GenerateStringConcatOperand(Expression expression)
+		{
+			if (expression is ConstantExpression constantExpression &&
```
davideicardi, line 2142: Now that you have `ToStringOrNull` function, I suspect that this `if` is not more useful. What happen if you remove it?

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -2090,23 +2115,39 @@ private MethodData FindBinaryOperator(string operatorName, Expression left, Expr
...
+		private static Expression ToStringOrNull(Expression expression)
+		{
+			var nullConstant = Expression.Constant(null);
+			Expression condition = Expression.Equal(
+				expression,
+				nullConstant);
+
+			return Expression.Condition(
+				condition,
+				Expression.ConvertChecked(nullConstant, typeof(String)),
```
davideicardi, line 2136: Maybe here you can use directly `nullConstant`? without any conversion? At the end `null` is a valid string....

davideicardi (changes requested):
Thank you!
Just two final small comments ...

src/DynamicExpresso.Core/Parsing/Parser.cs
davideicardi: I mean writing directly:
```
return Expression.Condition(
				condition,
				nullConstant,
				Expression.Call(expression, _toStringMethod));
```

It should be the equivalent of:
```
if (value == null) {
  return null;
}
else {
  return value.ToString();
}
```

src/DynamicExpresso.Core/Parsing/Parser.cs
nemo048: Hello @davideicardi 😄 You are right. The `if` is not more required. The tests work fine without it

src/DynamicExpresso.Core/Parsing/Parser.cs
nemo048: Unfortunately it is not possible 😞 I tried `nullConstant` without Type definition but `Expression.Condition` required that `then` and `else` if-branches should be of the same type.

nemo048:
Also i found one more issue in my code. We can't directly compare value types and `null`. For example `String_Concatenation_check_string_method` test failed without current changes. Now everything seems to be fine 😄

davideicardi (approved):
Thanks! Good work!

PR #153, opened 2021-06-07
No longer stop parsing an expression after invoking a custom function

### LINKED ISSUE #148: Global functions chained with linq causes parse exceptions from 2.5.0
So in these days, I've updated from 2.3.1 to 2.6.0 and some specific test cases in my own test suite started failing.

Upon bisecting it seems 2.5.0 is causing those test cases to fail so I made a test unit and tried it on both 2.4.1 and 2.5.0 and I can reliably see it failing due to a parse exception in 2.5.0 while it worked fine in 2.4.1.

From a c# standpoint, the syntax is correct, though.

This is the test case that can show the wrong behaviour and it seems likely related to #144 

```
		[Test]
		public void Func_Global_Invocation()
		{
			Func<object[], double, double, object[]> subArray = (entries, skipFirst, skipLast) => entries.Take(entries.Length - (int)skipLast).Skip((int)skipFirst).ToArray();

			target.SetVariable("arr1", new object[] { 1d, 2d, 3d });
			target.SetFunction("SubArray", subArray);

			Assert.AreEqual(2, target.Eval("SubArray(arr1, 1, 1).First()"));
		}
```

On 2.4.1 this works fine the eval returns 2 and the test succeeds. in 2.5.0 upward this will return a parse exception (for note this is a checkout of the current master to have a recollection of lines references)

```
  Message: 
    DynamicExpresso.Exceptions.ParseException : Syntax error (at index 20).
  Stack Trace: 
    Parser.ValidateToken(TokenId t, String errorMessage) line 2474
    Parser.Parse() line 68
    Parser.Parse(ParserArguments arguments) line 24
    Interpreter.ParseAsLambda(String expressionText, Type expressionType, Parameter[] parameters) line 443
    Interpreter.Parse(String expressionText, Type expressionType, Parameter[] parameters) line 343
    Interpreter.Eval(String expressionText, Type expressionType, Parameter[] parameters) line 420
    Interpreter.Eval(String expressionText, Parameter[] parameters) line 397
    OtherTests.Func_Global_Invocation() line 104
```

metoule:
Fix #148.

#138 introduced a regression when chaining method calls after the invocation of a custom function registered via `SetFunction` (e.g. `MyFunc().First()`): the parsing of the expression was stopped after the method call.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -514,9 +514,8 @@ namespace DynamicExpresso.Parsing
 						return ParseLambdaInvocation(lambda, tokenPos);
 
 					if (expr is MethodGroupExpression methodGroup)
-						return ParseMethodGroupInvocation(methodGroup, tokenPos);
-
-					if (typeof(Delegate).IsAssignableFrom(expr.Type))
+						expr = ParseMethodGroupInvocation(methodGroup, tokenPos);
+					else if (typeof(Delegate).IsAssignableFrom(expr.Type))
 						expr = ParseDelegateInvocation(expr, tokenPos);
 					else
 						throw CreateParseException(tokenPos, ErrorMessages.InvalidMethodCall, GetTypeName(expr.Type));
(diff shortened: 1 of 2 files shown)


davideicardi (approved):
Great!

PR #154, opened 2021-06-11
add dynamic unary, binary and conditional operations support

israellot:
This commit adds the ability to do unary, binary and conditional operations on dynamic types 
should address https://github.com/davideicardi/DynamicExpresso/issues/72


davideicardi (approved):
Thank you! Good work!

PR #156, opened 2021-06-18
Bind dynamic properties and methods of strongly typed classes

israellot:
This is a follow up to my last pull request. 
Dynamic properties and methods of strongly type classes are currently not being recognized as such because the resulting expression if of type Object instead of a dynamic expression. 
The proposed change is to treat expressions on boxed objects as dynamic, allowing late binding.
I've included a test case of a particular use case of mine that was failing otherwise.

### DIFF
--- a/src/DynamicExpresso.Core/InterpreterOptions.cs
+++ b/src/DynamicExpresso.Core/InterpreterOptions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace DynamicExpresso
 {
@@ -22,6 +22,10 @@ namespace DynamicExpresso
 		/// Variables and parameters names are case insensitive.
 		/// </summary>
 		CaseInsensitive = 8,
+		/// <summary>
+		/// Allow treating expressions of type Object as dynamic
+		/// </summary>		
+		LateBindObject = 16,
 		/// <summary>
 		/// Load all default configurations: PrimitiveTypes + SystemKeywords + CommonTypes
 		/// </summary>
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1384,9 +1384,11 @@ namespace DynamicExpresso.Parsing
 			return typeof(IDynamicMetaObjectProvider).IsAssignableFrom(type);
 		}
 
-		private static bool IsDynamicExpression(Expression instance)
+		private bool IsDynamicExpression(Expression instance)
 		{
-			return instance != null && instance.NodeType == ExpressionType.Dynamic;
+			return instance != null &&
+				(instance.NodeType == ExpressionType.Dynamic ||
+				(_arguments.Settings.LateBindObject && instance.Type == typeof(object)));
 		}
 
 		private static Type GetNonNullableType(Type type)
--- a/test/DynamicExpresso.UnitTest/DynamicTest.cs
+++ b/test/DynamicExpresso.UnitTest/DynamicTest.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.CSharp.RuntimeBinder;
+using Microsoft.CSharp.RuntimeBinder;
 using NUnit.Framework;
 using System;
 using System.Dynamic;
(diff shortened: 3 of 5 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1386,7 +1386,8 @@ private static bool IsDynamicType(Type type)
 
 		private static bool IsDynamicExpression(Expression instance)
 		{
-			return instance != null && instance.NodeType == ExpressionType.Dynamic;
+			return instance != null &&
+				(instance.NodeType == ExpressionType.Dynamic || instance.Type == typeof(object));
```
davideicardi, line 1390: There are no other ways to understand if the expression is declared as `dynamic`?
From outside just checking if it is an object seems to be wrong. But I admit that I don't see any real problem.

Maybe what I can just think of is what happen if you add some tests where we use an object? For example writing something like:
```
		public class NotDynamicProperties
        {
			public object Foo => 100;
			public object Bar() => 100;
        }

		[Test]
		public void Math_with_not_dynamic_properties_of_class()
		{
			var notDynamicProperties = new NotDynamicProperties();
			var del = interpreter.ParseAsDelegate<Func<NotDynamicProperties, int>>("d.Foo+d.Bar()", new[] { "d" });
			var r = del.Invoke(dynamicProperties);
			// TODO add an assert to check exception
		}
```

I would expect a ParseException. Correct?

test/DynamicExpresso.UnitTest/DynamicTest.cs
```diff
@@ -242,6 +242,26 @@ public void Math_with_dynamic_properties()
...
+		public void Math_with_dynamic_properties_of_class()
+		{
+
+			var dynamicProperties = new DynamicProperties();
+
+			var interpreter = new Interpreter();
+
+			var del = interpreter.ParseAsDelegate<Func<DynamicProperties, int>>("d.Foo+d.Bar()", new[] { "d" });
+			var r = del.Invoke(dynamicProperties);
+
```
davideicardi, line 261: Can you just add an assert to verify that the result is the one that we expect?
Can you also check the what happen if you just call `Eval`? What is the return type?
My concern is that in that case the return type will be object? Not that this is a real problem, but just to understand ...

davideicardi (changes requested):
Thanks! Added a couple of comments below...

src/DynamicExpresso.Core/Parsing/Parser.cs
israellot: This this is effective the same as the test I've included, since `public object Foo` and `public dynamic Foo` results in the same IL afaik. With the change proposed in this pull request, your suggested test won't throw.

test/DynamicExpresso.UnitTest/DynamicTest.cs
israellot: Sorry for the missing Assert. I've added tests for the conditions you mentioned. 
The return type of eval is Int32 as expected. The dynamic binding takes care of it as it should.

src/DynamicExpresso.Core/Parsing/Parser.cs
israellot: To be clear, in "real" C#, the dynamic keyword instructs the compiler to wire up the late biding code instead of the normal path which would simply not compile. Example : 
image
What this pull requests enable is late biding on objects without the dynamic keyword being required, which is a tiny deviation from C#. 
The real problem is that when doing reflection on a type like `DynamicProperties` , the properties and methods that are declared as `dynamic` are seen as having an `object` type. If we have an instance of that class, then we can get the actual type for those.

davideicardi: Yes, my concern is that this kind of "late binding" can introduce some bad side effect.

Maybe we can enable it using a feature flag? We can add a "ObjectLateBinding" option [here](https://github.com/davideicardi/DynamicExpresso/blob/master/src/DynamicExpresso.Core/InterpreterOptions.cs). Inside `IsDynamicExpression` check `|| instance.Type == typeof(object)` only if `ObjectLateBinding` is set.
And in the tests check the different behaviours depending on the flag.

What do you think?

israellot: That's probably wise and a very good idea. I'll amend the pull request with those.

israellot: Done @davideicardi ,
I've modified the test so it asserts an exception is throw when the flag is not enabled

src/DynamicExpresso.Core/InterpreterOptions.cs
```diff
@@ -25,6 +25,10 @@ public enum InterpreterOptions
 		/// <summary>
 		/// Load all default configurations: PrimitiveTypes + SystemKeywords + CommonTypes
 		/// </summary>
+		LateBindObject = 16,
+		/// <summary>
+		/// Allow treating expressions of type Object as dynamic
+		/// </summary>
```
davideicardi, line 31: I think that you should swap comments. The documentation should be on top ...

davideicardi (changes requested):
Thanks! One last small changes inside comments ...

src/DynamicExpresso.Core/InterpreterOptions.cs
israellot: Oh, my bad. Already done. Thanks!

davideicardi (approved):
Thank you! Good work!

Issue #162, opened 2021-10-05
Boolean operators not working for dynamic objects


AndiVarga:
Hello,

I've encountered this issue when trying to use boolean operators with dynamic properties:

```
dynamic dyn = new ExpandoObject();
dyn.val1 = 100;
dyn.val2 = 200;

var interpreter = new Interpreter();
interpreter.SetVariable("dyn", dyn);

interpreter.Eval("dyn.val1 == 100"); //returns true, as expected
interpreter.Eval("dyn.val1 == 100 && dyn.val2 == 200"); // throws exception: "Invalid argument value\r\nParameter name: operation
```

Please advise if this is indeed a bug or something is not properly configured.

davideicardi:
Hi @AndiVarga . Yes it looks like a bug.

You are working with the latest version?

Can you try to reproduce it in standard C# code? I would expect that it should compile, but just to be sure ...

davideicardi:
Yes, it looks like a bug.
As usual any help/PR is appreciated ;-) .

rotsin:
In 

```
private Expression ParseConditionalAnd()
{
	var left = ParseLogicalOr();
	while (_token.id == TokenId.DoubleAmphersand)
	{
		NextToken();
		var right = ParseLogicalOr();
		CheckAndPromoteOperands(typeof(ParseSignatures.ILogicalSignatures), ref left, ref right);
		left = GenerateBinary(ExpressionType.AndAlso, left, right);
	}
	return left;
}
```

which is called when there are dynamic expressions, expression type AndAlso is passed, problem is that Microsoft.CSharp.RuntimeBinder.Binder.BinaryOperation does not seem to support this type of operation. Changing it to And seems to resolve the issue, also added a small test to check that Andi's scenario works and it does, and all other tests run successfully.

Cound not push my branch to the repo (I may need rights), but the fix should be reasonably simple that there is no need to get my on the PR pipeline.

metoule:
Changing `AndAlso` to `And` would impact all operations that rely on the "short-circuiting" logical AND operator, which doesn't seem like a good idea. It would mean that all `&&` operations are considered as `&`, which will lead to unexpected side effects.

For example:

```` 
string test = null;
var res = test != null && test.IsNormalized();
````

using `AndAlso` ensures that the second operand is not evaluated, and therefore there's no `NullReferenceException`. Using `And` would lead to an unexcepted exception.

rotsin:
[...]

PR #163, opened 2021-10-06
Boolean operators not working for dynamic objects #162

rotsin:
Resolved an issue with AndAlso and OrElse not working with dynamic objects. Root cause of this was that the binary binder for dynamics explicitly does not support these two operations.

Code enhancements and guidance graciously provided by @metoule

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-﻿# Dynamic Expresso
+# Dynamic Expresso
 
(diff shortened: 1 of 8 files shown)


davideicardi:
@rotsin  Can you just update the branch? Sorry but I have merged a PR that is waiting for some time...

PR #186, opened 2021-11-14
Allow explicit return type object

### LINKED ISSUE #185: Bug when trying to upcast a delegate return type
```
var interpreter = new Interpreter().SetVariable("a", 123L);  
var del = interpreter.ParseAsDelegate<Func<object>>("a*2");
```

```
var interpreter = new Interpreter().SetVariable("a", 123L);  
var del = interpreter.ParseAsDelegate<Func<dynamic>>("a*2");
```
  
These simple examples used to run just fine on 2.8X and fails at 2.9.5.
I believe support to auto upcasting should be maintained.

metoule:
No longer omit the conversion expression when the return type is explicitly object.
Force the return type of a lambda expression to `typeof(void)` to prevent the emission of a conversion expression.

Fixes #185


davideicardi:
Thank you! I will look at this better in the following days.
It looks a little wrong to use `void` in that case, maybe we can find some alternatives ...

metoule:
I agree that it's a bit wrong to use void in that case.. I considered a few alternatives that didn't feel great either:
* creating a dummy type `DontEmitConversionExpression`
* add a new parser settings 

In the end, since this is purely internal for when we parse a lambda expression, it seemed acceptable to use void to minimize the impact. 
 
If you have a better idea, I'm all for it :)

davideicardi:
@metoule 
I'm doing some experiments with the following solution #190 . It is still in progress and not sure that this will actually solve the problem or it will be feasible. But I will try ...
If you already have any kind of feedbacks please let me know.

If in some days I will not be able to complete it we can fallback to your solution, that in anycase looks like to be good. Ok?

davideicardi (approved):
LGTM! Thank you as usual!

PR #189, opened 2021-11-15
Github workflows: build the solution only once

### LINKED ISSUE #161: Weird file version  (0.0.1.0) for .net standard package
Hi

I have a weird issue with the file version of the packages
I am using version 2.8.1, but when installed is file version 0.0.1.0
Any ideas why?
Strange thing is that when downgrading to 2.3.3 which i used to use, is see the same issue, and i am sure it wasnt like that before...

metoule:
Fixes #161


davideicardi (approved):
Nice! Good catch!

Issue #191, opened 2021-11-22
Parser.MethodHasPriority regression


waclaw66:
Method [Parser.MethodHasPriority](https://github.com/dynamicexpresso/DynamicExpresso/blob/3fd6251854f72258345871ec8e945e085e34e464/src/DynamicExpresso.Core/Parsing/Parser.cs#L2285) causes a regression in my tests. None of two possible methods is chosen.

methods:
```
0: {System.Collections.Generic.IEnumerable`1[System.Object] Select[TSource] System.Collections.Generic.IEnumerable`1[TSource], System.String)}
1: {System.Collections.Generic.IEnumerable`1[System.Object] Select(System.Collections.IEnumerable, System.String)}
```
and args passed:
```
0: {System.Collections.Generic.List`1[System.Object]}
1: {System.String}
```

See following test example...

```
namespace EvalTest
{
    using System.Collections;
    using System.Collections.Generic;
    using DynamicExpresso;
    using NUnit.Framework;

    [TestFixture]
    public class EvalTest
    {
        private Interpreter interpreter = new Interpreter();

        [OneTimeSetUp]
        public void Setup()
        {
            this.interpreter.Reference(typeof(Utils));
        }

        [Test]
        public void EvaluateTest()
        {
            var list = new[] { 1, 2, 3 };
            Assert.DoesNotThrow(() => { this.Evaluate("Utils.Array(1, 2, 3)"); });
            Assert.DoesNotThrow(() => { this.Evaluate("Utils.Select(Utils.Array(\"a\", \"b\"), \"x+x\")"); });
        }

        private object Evaluate(string expression, params Parameter[] parameters)
        {
            return this.interpreter.Parse(expression, parameters).Invoke(parameters);
        }
    }

    public class Utils
    {
        public static List<T> Array<T>(IEnumerable<T> collection)
        {
            return new List<T>(collection);
        }

        public static List<dynamic> Array(params dynamic[] array)
        {
            return Array((IEnumerable<dynamic>)array);
        }

        public static IEnumerable<dynamic> Select<TSource>(IEnumerable<TSource> collection, string expression)
        {
            return new List<dynamic>();
        }

        public static IEnumerable<dynamic> Select(IEnumerable collection, string expression)
        {
            return new List<dynamic>();
        }
    }
}
```

metoule:
The regression occurred in 2.9.0, so it's linked to PR #140 that adds support of lambda expressions.

[...]

PR #192, opened 2021-11-22
Allow a params parameter to have a generic type

metoule:
This is a fix for part of #191. The goal is to allow a `params` method parameter to be a generic. 
Use case:

```c#
private static class Utils
{
  public static List<T> Array<T>(params T[] array)
  {
    return new List<T>(array);
  }
}

var listInt = target.Eval<List<int>>("Utils.Array(1, 2, 3)");
Assert.AreEqual(new[] { 1, 2, 3 }, listInt);
```

Previously, the generic type was not promoted from the actual array type.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1943,7 +1943,15 @@ namespace DynamicExpresso.Parsing
 
 				if (paramsArrayTypeFound != null)
 				{
-					var promoted = PromoteExpression(currentArgument, paramsArrayTypeFound.GetElementType(), true);
+					var paramsArrayElementType = paramsArrayTypeFound.GetElementType();
+					if (paramsArrayElementType.IsGenericParameter)
+					{
+						paramsArrayPromotedArgument = paramsArrayPromotedArgument ?? new List<Expression>();
+						paramsArrayPromotedArgument.Add(currentArgument);
+						continue;
+					}
+
+					var promoted = PromoteExpression(currentArgument, paramsArrayElementType, true);
 					if (promoted != null)
 					{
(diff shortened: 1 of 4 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1961,6 +1969,16 @@ private static bool CheckIfMethodIsApplicableAndPrepareIt(MethodData method, Exp
 				var paramsArrayElementType = paramsArrayTypeFound.GetElementType();
 				if (paramsArrayElementType == null)
 					throw new Exception("Type is not an array, element not found");
+
+				if (paramsArrayElementType.IsGenericParameter)
+				{
+					var actualTypes = paramsArrayPromotedArgument.Select(_ => _.Type).Distinct().ToArray();
+					if (actualTypes.Length != 1)
+						throw new Exception($"The type arguments for method '{method.MethodBase}' cannot be inferred from the usage.");
```
metoule, line 1977: I would have preferred to throw a `ParserException` instead of the current `Exception`, but it's really far down the stack, and that would require passing the token position all the way down, which didn't seem great.

davideicardi (approved):
LGTM. Thank you!

src/DynamicExpresso.Core/Parsing/Parser.cs
davideicardi: @metoule 
Maybe it is better to throw a `ParseException` anyway with a -1 position? Just because in this way a caller can catch only it and avoid using a generic `Exception`. Here and also some line above ...

What do you think?

metoule: That's a good idea, I'll update the PR! Thanks!

PR #195, opened 2021-11-25
Add unit tests for lambda parse exception and and to check return type

davideicardi:
Just added 2 unit tests for lambda expressions. They could be useful to avoid some regression in the future.


metoule (approved):
LGTM!

PR #199, opened 2021-12-07
Added support of param arrays within constructors

### LINKED ISSUE #196: Support for params within class constructor
Lets have following class...

```c#
public class Array<T> : List<T>
{
    public Array() {}
    public Array(IEnumerable<T> collection) : base(collection) {}
    public Array(params T[] items) : base(items) {}
}
```

It would be nice to have support for such construction...
`new Array<int>(1, 2, 3)`

`Parser.ParseNew()` method doesn't handle it properly.
That construction can be achieved only with static method currently.

metoule:
This simply re-uses the existing method resolution, but for constructors (instead of a special resolution).

Fix #196


davideicardi (approved):
LGTM! Thank you!

PR #201, opened 2021-12-08
Allow a lambda expression to access the initial expression's parameters

### LINKED ISSUE #200: Parameters are not passed to lambda expression
Following example returns... `Unknown identifier 'x' (at index 5).`
``` c#
var listInt = target.Eval<List<int>>("list.Select(n => n > x)", new Parameter("list", new[] { 1, 2, 3 }), new Parameter("x", 1));
Assert.AreEqual(new[] { 2, 3 }, listInt);
```

metoule:
Fix #200

It works by making all the initial expression parameters as variables of the lambda expression interpreter.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/ParserSettings.cs
+++ b/src/DynamicExpresso.Core/Parsing/ParserSettings.cs
@@ -6,11 +6,11 @@ namespace DynamicExpresso.Parsing
 {
 	internal class ParserSettings
 	{
-		private Dictionary<string, Identifier> _identifiers;
-		private Dictionary<string, ReferenceType> _knownTypes;
-		private HashSet<MethodInfo> _extensionMethods;
+		private readonly Dictionary<string, Identifier> _identifiers;
+		private readonly Dictionary<string, ReferenceType> _knownTypes;
+		private readonly HashSet<MethodInfo> _extensionMethods;
 
-		public ParserSettings(bool caseInsensitive,bool lateBindObject)
+		public ParserSettings(bool caseInsensitive, bool lateBindObject)
 		{
 			CaseInsensitive = caseInsensitive;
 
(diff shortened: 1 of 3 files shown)


src/DynamicExpresso.Core/Parsing/ParserSettings.cs
```diff
@@ -33,6 +33,19 @@ public ParserSettings(bool caseInsensitive,bool lateBindObject)
 			LambdaExpressions = false;
 		}
 
+		/// <summary>
+		/// Creates a deep copy of the current settings, so that the identifiers/types/methods can be changed
+		/// without impacting the existing settings.
+		/// </summary>
+		internal ParserSettings Clone()
```
davideicardi, line 40: I usually implement clone method using the "copy constructor" pattern.
I think it is easier and we can leave readonly on the variables.

Essentially we can have an additional constructor like:

```
private ParserSettings(ParserSettings other)
{
// TODO Here we can copy the variables
}
```

And then the clone method can be implemented as:

```
internal ParserSettings Clone()
{
return new ParserSettings(this);
}
```

We can also eval to put `Clone` as public. What do you think?

davideicardi (approved):
Approved with a suggestion.

src/DynamicExpresso.Core/Parsing/ParserSettings.cs
metoule: The reason I went without a copy constructor is to be able to use the `MemberwiseClone()` method, which creates a shallow copy of all fields. Then, I can new instances of the various lists when needed. My main fear with a copy constructor is always to forget to update it when a new field is added..

I don't have a strong opinion either way, so I'll implement your suggestion and make it public :)

davideicardi: Yes, I get your point of view. My fealing is that removing `readonly` can introduce more strange bugs. But also for me this is not a strong opinion. As you prefer.

metoule: I agree with your assessment that removing the `readonly` modifier can lead to strange bugs, so I added the copy ctor.

davideicardi (approved):

PR #208, opened 2021-12-11
Improve bitwise operations support

### LINKED ISSUE #206: Does it support bitwise operations
Hi

   I like this prj and want to use it for our iot applications

Our user cases involve bitwise operations which I am not sure are supported or not.

Regards

Ygl

metoule:
Add support of the unary bitwise complement operator (~)
Add support of the shift operators (>> and <<)
Fix #206

```c#
var x = 0b_1100_1001_0000_0000_0000_0000_0001_0001;
target.SetVariable("x", x);

Assert.Equal(~x, target.Eval<uint>("~x"));
Assert.Equal(x >> 4, target.Eval<uint>("x >> 4"));
Assert.Equal(x << 4, target.Eval<uint>("x << 4"));
```

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -513,6 +513,11 @@ namespace DynamicExpresso.Parsing
 			return left;
 		}
 
+		/// <summary>
+		/// Returns true if and only if the current token is in fact a shift operator.
+		/// In that case <paramref name="shiftType"/> is set to the proper expression type.
+		/// If the function returns false, <paramref name="shiftType"/> shouldn't be used.
+		/// </summary>
 		public bool IsShiftOperator(out ExpressionType shiftType)
 		{
 			// >> is not a token, because it conflicts with generics such as List<List<int>>
@@ -530,7 +535,8 @@ namespace DynamicExpresso.Parsing
 				return true;
 			}
 
-			shiftType = ExpressionType.DebugInfo; // dummy expression type
+			// dummy expression type that shouldn't be used
+			shiftType = ExpressionType.DebugInfo;
 			return false;
 		}
 
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -355,7 +355,7 @@ namespace DynamicExpresso.Parsing
 			{
 				var op = _token;
 				NextToken();
-				var right = ParseAdditive();
+				var right = ParseShift();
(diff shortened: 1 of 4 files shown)


metoule:
Note that I didn't add new tokens for the shift operators `>>` and `<<` because it would prevent the proper parsing of generic type definitions such as `List<List<int>>`.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -499,6 +499,41 @@ private Expression ParseAdditive()
...
+			}
+			// << could be a token, but is not for symmetry
+			else if (_token.id == TokenId.LessThan && _parseChar == '<')
+			{
+				NextToken(); // consume next < 
+				shiftType = ExpressionType.LeftShift;
+				return true;
+			}
+
+			shiftType = ExpressionType.DebugInfo; // dummy expression type
```
davideicardi, line 533: Maybe int shi case we can set it to `null`? I don't like using null but in this case it should be safe and more clear I think.

davideicardi (approved):
Very good! Just a small comment.

src/DynamicExpresso.Core/Parsing/Parser.cs
metoule: `shiftType` is an enum, so it can't be assigned to null. I could change the signature to use a nullable enum value, but that would require the caller to call the `Value` property, which doesn't seem ideal. In any case, if the method returns false, the `out` value should not be used. 

I'll add more comments to explain why it's not set to null.

davideicardi: Ah yes, sorry. You are right. Ok for me.

waclaw66:
Could you please add directives to unit tests for language versions that do not support it yet?

metoule:
Which language versions don't support it? The tests ran successfully for .NET Framework 4.5 and 4.6.1 and .NET Core 3.1 and 5.0.

waclaw66:
obrazek

metoule:
That's really strange - which version of Visual Studio are you using? It should work out of the box from VS2019 onwards.
If you can't upgrade Visual Studio, that's not a huge deal: it's just syntaxic sugar to declare an integer from its binary representation. 

You can replace replace that variable declaration with any integer, for example `var x = 123`, and the test will be the same.

For the other tests, 

```c#
var a = 0b_1111_1000;
var b = 0b_0001_1100;
```

can be replaced with

```c#
var a = 248;
var b = 28;
```

Be careful for test `Bitwise_operations_uint_int`, the variable `x` must be an unsigned integer `var x = 248u`.

metoule:
Or you can install VS community edition :)

waclaw66:
I use VS2017, that is the problem :D

PR #209, opened 2021-12-11
Improve support of nullable types

### LINKED ISSUE #205: InvalidOperationException for Nullable types
**1 Having the following expression:**

```c#
var interpreter = new Interpreter()
        .SetVariable("date1", DateTimeOffset.UtcNow)
        .SetVariable<DateTimeOffset?>("date2", null);

interpreter.Eval("date1 - date2");
```

Will throw an exception

```c#
System.InvalidOperationException: The binary operator Subtract is not defined for the types 'System.DateTimeOffset' and 'System.Nullable`1[System.DateTimeOffset]'.
```

**2 Also this expression**

```c#
interpreter.Eval("date2?.Day");
```

will throw

```c#
DynamicExpresso.Exceptions.ParseException: No property or field 'Day' exists in type 'DateTimeOffset?' (at index 7).
```

Eventually we also need to verify one more expression once above is fixed. The following expression must return true or false bool value.

```c#
interpreter.Eval("(date1 - date2)?.Days < 7");
```

metoule:
Promote all operands of binary operators.
When applying the ?. operator to nullable types, emit the member access on the underlying type.
Fix #205

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -403,12 +403,6 @@ namespace DynamicExpresso.Parsing
 				//			op.text, ref left, ref right, op.pos);
 				//}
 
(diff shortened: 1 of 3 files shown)


davideicardi (approved):
LGTM!

PR #210, opened 2021-12-12
MethodHasPriority: improve resolution

### LINKED ISSUE #203: IEnumerable regression after lambda support
Lets have following example...

``` c#
private static class Utils
{
    public static List<T> Array<T>(IEnumerable<T> collection)
    {
        return new List<T>(collection);
    }

    public static List<dynamic> Array(params dynamic[] array)
    {
        return Array((IEnumerable<dynamic>)array);
    }
}

var listInt = target.Eval<List<int>>("Utils.Array(list)", new Parameter("list", new[] { 1, 2, 3 }));
Assert.AreEqual(new[] { 1, 2, 3 }, listInt);
```

it uses `Array<T>(params T[] array)` instead of `Array<T>(IEnumerable<T> collection)` which leads to enumerable missintepretation.

metoule:
MethodHasPriority: use promoted parameters when the method's argument type is an inferred generic type.
Fixes #203

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -2340,11 +2340,11 @@ namespace DynamicExpresso.Parsing
 				var methodParamType = GetParameterType(methodParam);
 				var otherMethodParamType = GetParameterType(otherMethodParam);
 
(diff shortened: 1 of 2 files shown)


davideicardi (approved):
Again LGTM! Thank you!

Issue #207, opened 2021-12-11
Unclear SetExpression with parameter behavior 


halamah:
Hi,

```c#
var interpreter = new Interpreter();
var parameter = new Parameter("x", typeof(int));
var expression = interpreter.Parse("x + 1", parameter).Expression;

var lambda = interpreter
    .SetExpression("value", expression)
    .Parse("value + 1", parameter);

var result = lambda.Invoke(1);
```

This will throw 

```
System.InvalidOperationException: variable 'x' of type 'System.Int32' referenced from scope '', but it is not defined
```

Trying to register identifier like below leads to same exception

```c#
var lambda = interpreter
    .SetExpression("value", expression)
    .SetIdentifier(new Identifier("x", parameter.Expression)) 
    .Parse("value + 1", parameter);
```

And only if I mention the identifier in primary expression `interpreter.Parse("x + value + 1", parameter)` will parse with success.

Could you please advise how to propagate `x` parameter to internal expression without mentioning it in primary one?

Thanks

halamah:
I assume this is because of UsedParameters are actually passed when creating Lambda.

So what the point of UsedParameters?

halamah:
It looks like the DynamicExpresso.Lambda wrapper is intended to simplify the usage, but on the other hand introduces some limitations.

For example in mentioned scenario DynamicExpresso.Lambda constructor successfully creates an Expression itself but then fails to create Lambda and even though an Expression itself successfully constructed I can not use it because of thrown exception.

Maybe I am missing something, but it would more flexible to return raw expression result without trying to compile it leaving up to a client to decide whether there is a need to compile an expression or use it as is.

Interpreter.ParseAsExpression looks more or less like what I described above (although in dynamic context it could be complicated to get a Delegate type), but even in this case it **first compiles constructed Expression with default behaviour using UsedParameters and then recompile once again as per passed Delegate type**.

Could you perhaps consider having a Interpreter.Parse which return DynamicExpresso.Lambda consisting only Expression, DeclaredParameters, UsedParameter and maybe some other properties, but no more methods. The same applies to Interpreter - only a single Parse method. 

Then all the lambda compile variations move to Extension methods over DynamicExpresso.Lambda. This approach will simplify further library extension, will separate parse and compile logic and moreover, if something is not implemented in the library, a client could take that raw lambda expression and write an extension on its own (currently all this stuff is internal).

Does it make sense?

davideicardi:
[...]

PR #211, opened 2021-12-12
Unclear SetExpression with parameter behavior #207

halamah:
A proposed refactored sample version of DynamicExpresso in terms of Parse and Compile separation.


halamah:
I had to comment out some tests related to Invoke with named parameters, like omitting the unused parameter or passing out of order parameters etc.

Also haven't found out the cause of failing tests related to LambdaExpression option. `target.Parse("list.Select(str => str.Length)")` - `str` here if of type Object.

davideicardi:
Thank you @halamah . I will study your changes and investigate about failing tests.

Can you sycnronize your branch with master (rebase) so that we always compare it with the latest version?

cc @metoule Any feedback on this?

halamah:
I've changed my PR as the initial one was really too cumbersome. Here I've just copy/pasted the interpreter to a new ExpressionInterpreter class with relevant changes. All these is fully backward compatible since the Interpreter class wasn't changed.

So having this new ExpressionInterpter + InterpreterExtensions I can do basically the same, but the client has more obvious control over produced Expression.

Plus I've just added a simple extension to Eval expression without even touching an Interpreter class.

```c#
var interpreter = new ExpressionInterpreter();
var result = interpreter.Eval("x + 1", x => 2);
Assert.Equals(result, 3);
```

src/DynamicExpresso.Core/ParseResult.cs
```diff
@@ -0,0 +1,67 @@
...
+	{
+		public ParseResult(
+			Expression expression,
+			IEnumerable<ParameterExpression> usedParameters,
+			IEnumerable<ParameterExpression> declaredParameters,
+			IEnumerable<ReferenceType> types,
+			IEnumerable<Identifier> identifiers) : base(expression, usedParameters, declaredParameters, types, identifiers)
+		{
+		}
+	}
```
metoule, line 66: Why do you have a generic version of `ParseResult`? It seems you don't use the generic parameter?

src/DynamicExpresso.Core/Parameter.cs
```diff
@@ -38,6 +38,11 @@ public Parameter(string name, Type type, object value = null)
 			Expression = System.Linq.Expressions.Expression.Parameter(type, name);
 		}
 
+		public static Parameter Is<T>(string name, T value)
+		{
+			return new Parameter(name, typeof(T), value);
+		}
+
```
metoule, line 45: Maybe rename it to `Create`, or `Make`, or `New`? I'd expect `Is` to return a boolean

src/DynamicExpresso.Core/ParseResult.cs
halamah: This generic ParseResult<> will store Type information and there is a generic Compile extension over ParseResult<> to Compile to target delegate.

You could perhaps Compile over basic ParseResult, but the intention was to Parse with proper ReturnType and then just compile to get current delegate without passing a DelegateType.

metoule: Ah, I see it now, thanks for the explanation. Good idea 👍

halamah:
I've tried both suggestions and

1) There will be name collisions for Parse method (new one return different type) and I don't really like to rename it.

2) Base class actually turns into a wrapper over Settings and Visitors - and this will work, but the intention is to mark current Interpreter and Lambda classes as Obsolete and don't support those further. All new changes will come to new ExpressionInterpreter and Extension methods only.

src/DynamicExpresso.Core/Parameter.cs
halamah: Changed to Create

halamah:
Yes, it could. But I'm note quite sure this will be "pretty" or even feasible. Current Lambda implementation stores compiled result and calling Invoke uses that already compiled lambda. Also there is a trick with UsedParameters - the issue this PR comes from (just to remind if a SetExpression text contains input parameter and a parent one doesn't - it won't be propagated).

On the other hand ParseResult is a raw parsed Expression (which is the most valuable part of the library) that could be used in numerous ways without compilation. Later if one needs to invoke the expression he must to compile it first.

Maybe there is some way to do what you suggest, but nothing comes to my mind so far.

I'm not sure, but for example `Interpreter.ParseAsExpression` contains this code

```c#
var returnType = delegateInfo.ReturnType;
if (returnType == typeof(object))
	returnType = typeof(void);
```

looks like a good optimisation, but from the other hand, why do we need the conversion at all inside Parser? Maybe the conversion should be moved out and up to client to decide whether he needs the conversion or not, because there are as many ways to use the library as many clients, and it would be very difficult trying to cover all of them. You could introduce an extension method to simplify the usage though.

Also this Lambda method

```c#
internal LambdaExpression LambdaExpression(Type delegateType)
{
	var types = delegateType.GetGenericArguments();
	// return type
	types[types.Length - 1] = _expression.Type;
	var genericType = delegateType.GetGenericTypeDefinition();
	var inferredDelegateType = genericType.MakeGenericType(types);
	return Expression.Lambda(inferredDelegateType, _expression, DeclaredParameters.Select(p => p.Expression).ToArray());
}
```

I'm not sure why is that for and will it work if delegate type is not a Func<> and also in this case it uses DeclaredParameters instead of UsedParameters, so many internal stuff that is not obvious outside.

Anyway my PR is just a suggestion and sample, two mentioned parts of code are omitted in my extension methods

davideicardi:
> On the other hand ParseResult is a raw parsed Expression (which is the most valuable part of the library) that could be used in numerous ways without compilation. Later if one needs to invoke the expression he must to compile it first.

@halamah  I completely agree on that sentence. It would be nice to split up the parse and compile phase.
I'm not sure how to do it in the best way but we this is a good start.

metoule:
Yes, `ParseResult` is clearly a step in the right direction! I believe the `Lambda` class can be made to encapsulate a `ParseResult`, while retaining the old methods. That way, it will be possible to add new methods that return a `ParseResult`, while still keeping the old way of doing things to avoid breaking changes.

I checked your PR locally to see if I can find a way to do that :)

metoule:
@halamah  I couldn't find a way to update this PR, so I created a new one based on this: #216

PR #213, opened 2021-12-14
Prototype to capture variables so that they can be used in an inner lambda

### LINKED ISSUE #212: Parsing Lambda expression pre-evalutes it with given arguments
When Parsing lambda expression, it evaluates the predicate and cannot be reused (e.g. from cache) with another parameter value.
See an example that demonstrates it.

``` c#
var list = new Parameter("list", new[] { 1, 2, 3 });
var value1 = new Parameter("value", 1);
var value2 = new Parameter("value", 2);
var expression = "list.Where(x => x > value)";
var lambda = target.Parse(expression, list, value1);
var result = lambda.Invoke(list, value2);
Assert.AreEqual(new[] { 3 }, result);
```

metoule:
The idea is:

- use a dictionary to store captured variables (= parameters from all levels of lambda expressions)
- when parsing an expression, first assign the captured parameter to the dictionary

For example, it we parse `x + 5`, and call `f(2)` we convert it to:

```c#
dict["x"] = 2;
return dict["x"] + 5;
```

Fix #212

### DIFF
--- a/src/DynamicExpresso.Core/Interpreter.cs
+++ b/src/DynamicExpresso.Core/Interpreter.cs
@@ -403,21 +403,6 @@ namespace DynamicExpresso
 			return lambda.LambdaExpression<TDelegate>();
 		}
 
-		internal LambdaExpression ParseAsExpression(Type delegateType, string expressionText, params string[] parametersNames)
-		{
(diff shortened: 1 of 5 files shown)


metoule:
Superseded by #257

PR #215, opened 2021-12-18
Fix incorrect overload resolution

### LINKED ISSUE #214: Incorrect resolution with methods with params[] array
This is a regression due to my PR #193. When there are two overloads that have the same arguments except for the params array, calling the method without the array leads to the improper method to be chosen.

```c#
internal static class Utils
{
  public static int WithParamsArray2(string str, Exception e) => 6;
  public static int WithParamsArray2(string str, Exception e, params string[] args) => 7;
}

var str = "str";
var e = new Exception();
target.SetVariable("str", str);
target.SetVariable("e", e);

// incorrectly uses the overload with the params array
Assert.AreEqual(6, target.Eval("Utils.WithParamsArray2(str, e)"));
Assert.AreEqual(7, target.Eval("Utils.WithParamsArray2(str, e, str, str)"));
```

metoule:
Fix incorrect overload resolution when two overloads differ only by params array arguments.

Fixes #214

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -2068,8 +2068,13 @@ namespace DynamicExpresso.Parsing
 			{
 				if (x.HasDefaultValue)
 					return Expression.Constant(x.DefaultValue, x.ParameterType);
+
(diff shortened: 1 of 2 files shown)


davideicardi (approved):
LGTM!

PR #219, opened 2021-12-23
Lazy compile in lambda

halamah:
Lazy compile delegate in Lambda so the internal Expression can be processed in custom way without default compile overhead.


src/DynamicExpresso.Core/Lambda.cs
```diff
@@ -14,28 +14,22 @@ public class Lambda
...
+			// Note: I always lazy compile the generic lambda. Maybe in the future this can be a setting because if I generate a typed delegate this compilation is not required.
+			_delegate = new Lazy<Delegate>(() =>
+				Expression.Lambda(_expression, _parserArguments.UsedParameters.Select(p => p.Expression).ToArray()).Compile());
 		}
 
 		public Expression Expression { get { return _expression; } }
 		public bool CaseInsensitive { get { return _parserArguments.Settings.CaseInsensitive; } }
 		public string ExpressionText { get { return _parserArguments.ExpressionText; } }
-		public Type ReturnType { get { return _delegate.Method.ReturnType; } }
+		public Type ReturnType { get { return Expression.Type; } }
```
halamah, line 32: @metoule @davideicardi Is it Ok to use Expression.Type as a ReturnType without compile to delegate?

halamah:
@davideicardi @metoule As a workaround to fix #207 could we merge in this lazy delegate compile, so I can use raw expression without default delegate compile overhead?

src/DynamicExpresso.Core/Lambda.cs
metoule: The delegate's return type should be the same as the expression type, because the delegate is built based on the expression type.

metoule:
That's a good first step, that has minimal impact on the current interface. However, that alone won't fix #207 because the root cause is the UsedParameters vs DeclaredParameters.
Maybe you can also backport the DetectUseParameters option?

I think maybe it should be set to false by default but that is potentially a major change. 

PS. I'm currently on holidays, with minimal access to a computer, so I won't be able to push commits for a few days.

halamah:
I won't use Lambda.Invoke. Since the exception highlighted in #207 comes from compile in Lambda constructor you can do the following:

```c#
// parse
var lambda = interpreter.SetExpression(innerExpression).Parse(expressionText);

// custom compile and pass all required parameters myself
var delegate = Expression.Lambda(lambda.Expression, ....).Compile();

// invoke 
var result = delegate.DynamicInvoke(....)
```

Am I missing something?

metoule:
In that case, the lazy compilation is enough :)

metoule (approved):
LGTM!

davideicardi (approved):
LGTM! Thank you very much @halamah !

I will merge this and release a new version.
But let's keep #211 and related PRs open, because I think that your original idea is the right solution for the future.

davideicardi:
Published v2.10.0! Thank you again @halamah

PR #225, opened 2022-01-31
Change project target frameworks

### LINKED ISSUE #224: .NET Framework 4.5.2 end of support is April 26, 2022
DynamicExpresso is configured for multi-targeting:

```xml
<TargetFrameworks>netstandard2.0;net461;net45</TargetFrameworks>
```

.NET 4.5 is reaching [end of support on April 26, 2022](https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework):

image

That means that the lowest supported version of .NET Framework is now .NET 4.6.2, which is [compatible with `netstandard2.0`](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0).

Therefore, I suggest the following: 
* only target `netstandard2.0` in DynamicExpresso.Core
* target `netcoreapp5.0;netcoreapp3.1;net462` in DynamicExpresso.UnitTest

metoule:
Fix #224

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![NuGet version](https://badge.fury.io/nu/DynamicExpresso.Core.svg)](http://badge.fury.io/nu/DynamicExpresso.Core)
 [![.NET CI](https://github.com/dynamicexpresso/DynamicExpresso/actions/workflows/ci.yml/badge.svg)](https://github.com/dynamicexpresso/DynamicExpresso/actions/workflows/ci.yml)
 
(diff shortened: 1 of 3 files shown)


davideicardi (approved):
LGTM!

davideicardi:
@metoule  Do you think we should really consider this as a breaking change? We will "break" only user on .NET 4.5. Ok technically speaking is a breaking change, but maybe we can just put a "BREAKING CHANGE" note in the release note.

metoule:
Now that I think about it, Nuget should handle this for us: it will stop suggesting updates to users on .NET4.5, but they can still use the 2.10.0 version.
I agree with your suggestion to put a note in the release notes :)

PR #227, opened 2022-02-03
The identifier detection now supports lambda parameters

### LINKED ISSUE #226: Parse unknown identifiers for lambda expressions
This is a very helpful feature, but when used against expressions with lambdas e.g. `list.Any(x => x == null)` - `x` here is an unknown identifier.

So I'm note sure whether this is an expected behaviour, because from expression inputs perspective - yes it's unknown, however it won't raise any compilation errors (since I enabled lambdas) and it's more likely a runtime variable rather than an unknown input identifier.

metoule:
I've added a regex to detect lambda parameters.

Fixes #226


davideicardi (approved):
LGTM! Thank you!

PR #229, opened 2022-02-19
Change error message when throw when same functions are registered

### LINKED ISSUE #159: ParseException with ArgsIncompatibleWithDelegate message is throw when multiple functions are registered with the same signature
We should throw instead a `ParseException` with `AmbiguousDelegateInvocation` that describe better the use case.

See #157 and #158 .

> in FindBestMethod, if there are more than one applicable method, the candidates are filtered by the MethodHasPriority method. If no method has priority (like in this particular case), then FindBestMethod returns an empty array, and therefore the caller can't see that there's in fact an ambiguity

See also [this explanation](https://github.com/dynamicexpresso/DynamicExpresso/pull/229#discussion_r815301079)

sagilio:
Fixes #159 
1. Change error message according to #159 
2. Public `FunctionIdentifier` to allow use replace function directly instead of add overload.

@davideicardi @metoule

### DIFF
--- a/src/DynamicExpresso.Core/Identifier.cs
+++ b/src/DynamicExpresso.Core/Identifier.cs
@@ -22,13 +22,13 @@ namespace DynamicExpresso
 		}
 	}
 
-	internal class FunctionIdentifier : Identifier
+	public class FunctionIdentifier : Identifier
 	{
-		internal FunctionIdentifier(string name, Delegate value) : base(name, new MethodGroupExpression(value))
+		public FunctionIdentifier(string name, Delegate value) : base(name, new MethodGroupExpression(value))
 		{
 		}
 
-		internal void AddOverload(Delegate overload)
+		public void AddOverload(Delegate overload)
 		{
 			((MethodGroupExpression)Expression).AddOverload(overload);
 		}
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -524,7 +524,7 @@ namespace DynamicExpresso.Parsing
 			// << could be a token, but is not for symmetry
 			else if (_token.id == TokenId.LessThan && _parseChar == '<')
 			{
-				NextToken(); // consume next < 
+				NextToken(); // consume next <
 				shiftType = ExpressionType.LeftShift;
 				return true;
 			}
@@ -1019,7 +1019,7 @@ namespace DynamicExpresso.Parsing
 			{
 				return ParseTypeKeyword(knownType);
 			}
-			
+
 			// Working context implementation
 			//if (it != null)
 			//    return ParseMemberAccess(null, it);
@@ -1228,7 +1228,7 @@ namespace DynamicExpresso.Parsing
 				applicableMethods = FindBestMethod(candidates.Select(_ => _.Method), args);
 
 			if (applicableMethods.Length == 0)
-				throw CreateParseException(errorPos, ErrorMessages.ArgsIncompatibleWithDelegate);
+				throw CreateParseException(errorPos, ErrorMessages.AmbiguousDelegateInvocation);
 
 			if (applicableMethods.Length > 1)
 				throw CreateParseException(errorPos, ErrorMessages.AmbiguousDelegateInvocation);
@@ -1250,7 +1250,7 @@ namespace DynamicExpresso.Parsing
 
 		private bool TryParseKnownType(string name, out Type type)
 		{
-			// if the type is unknown, we need to restart parsing 
+			// if the type is unknown, we need to restart parsing
 			var originalPos = _token.pos;
 			_arguments.TryGetKnownType(name, out type);
 
@@ -2362,7 +2362,7 @@ namespace DynamicExpresso.Parsing
 					}
 				case ExpressionType.Parameter:
 					return true;
-				
+
 			}
 
 			return false;
@@ -2497,7 +2497,7 @@ namespace DynamicExpresso.Parsing
 			return GenerateBinary(ExpressionType.GreaterThan, left, right);
 		}
 
-		
+
 
 		private Expression GenerateGreaterThanEqual(Expression left, Expression right)
 		{
@@ -3090,7 +3090,7 @@ namespace DynamicExpresso.Parsing
 		}
 
 		/// <summary>
-		/// Expression that wraps over an interpreter. This is used when parsing a lambda expression 
+		/// Expression that wraps over an interpreter. This is used when parsing a lambda expression
 		/// definition, because we don't know the parameters type before resolution.
 		/// </summary>
 		private class InterpreterExpression : Expression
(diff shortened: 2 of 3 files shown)


src/DynamicExpresso.Core/Identifier.cs
```diff
@@ -22,13 +22,13 @@ public Identifier(string name, Expression expression)
 		}
 	}
 
-	internal class FunctionIdentifier : Identifier
+	public class FunctionIdentifier : Identifier
```
davideicardi, line 25: I'm not sure if making `FunctionIdentifier` public is a good idea.
For 2 main reasons:
1. calling `AddOverload` or other functions that change the class state can alter the state of the overall `Interpreter` in unexpected way. See https://github.com/dynamicexpresso/DynamicExpresso#performance-and-multithreading
2. having a wider public API makes refactoring and future extensions more difficult to avoid breaking changes

If there aren't specific needs I would prefer to leave it `private`.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1228,7 +1228,7 @@ private Expression ParseMethodGroupInvocation(MethodGroupExpression methodGroup,
 				applicableMethods = FindBestMethod(candidates.Select(_ => _.Method), args);
 
 			if (applicableMethods.Length == 0)
```
davideicardi, line 1230: Given the issue description I thought that we should throw `AmbiguousDelegateInvocation` when  `applicableMethods.Length > 1` . Why we throw when `== 0`?
Maybe I missing something .

@metoule  maybe you can help us ;-)

test/DynamicExpresso.UnitTest/GithubIssues.cs
```diff
@@ -221,11 +221,40 @@ static bool GetGFunction2(string arg = null)
...
+			GFunction gFunc1 = GetGFunction1;
+			GFunction gFunc2 = GetGFunction2;
+
+			var interpreter = new Interpreter();
+			interpreter.SetFunction("GFunction", gFunc1);
+			interpreter.SetFunction("GFunction", gFunc2);
+			interpreter.SetVariable("arg", "arg");
+
+			// ambiguous call
+			var exception = Assert.Throws<ParseException>(() => interpreter.Eval("GFunction(arg)"));
```
davideicardi, line 247: What happen with standard C# when we have two functions like these?
I would expect to have the same behavior with `SetFunction` and `SetIdentifier` ...

davideicardi (commented):
Thank you for your help @sagilio !

I'm just not sure about some details. See my comments.

src/DynamicExpresso.Core/Identifier.cs
sagilio: You are right, I also think it is not the best way. I only want to replace an existing function identifier instead of adding a new overload.

test/DynamicExpresso.UnitTest/GithubIssues.cs
sagilio: There is not an overload relation between `gFunc1` with `gFunc2`, I think it should replace the `gFunc1` delegate when adding the `gFunc2`

src/DynamicExpresso.Core/Identifier.cs
davideicardi: So you need to use an existing `Interpreter` instance but with a different function? Right?
In this case maybe the best way is to simply create a new `Interpreter`. You can for example create a factory to create your Interpreter instances and in one case use the new function.

What do you think?

sagilio: If the existing `Interpreter` has a lot of functions and variables, I need to re-add them when I create a new `Interpreter` for the new function.  And the new function is not very "different", it has the same signature as the old one. It's just wrongly regarded as overload when calling the `SetFunction`, and then the `Interpreter` will can not parse any expression text with that function name.

Now version, `SetIdentifier`, `SetVariable`, and `SetExpression` all will replace the existing expression in the map when setting the same name, but `SetFunction` not.

davideicardi: Ok, this is a good point. I agree that this was probably an error, `SetFunction` should replace the old function and not add an overload.
My suggestion is to align the `SetFunction` behaviour with the previous `SetXYZ` functions, replacing the current value, and add another function called `SetFunctionOverload` that instead add an overload if the function already exist.
Unfortunately this will be a breaking change ... but ok, not a big problem, I will create a new major version.

@metoule  What do you think?

src/DynamicExpresso.Core/Parsing/Parser.cs
metoule: That's precisely the issue I highlighted, and why it's actually quite difficult to fix properly :D 
The issue resides in the `FindBestMethod` method. When there are multiple applicable overloads, they're filtered thanks to the `MethodHasPriority` method:

```c#
private static MethodData[] FindBestMethod(IEnumerable<MethodData> methods, Expression[] args)
{
  // empty array if there are no applicable methods
  var applicable = methods
    .Where(m => CheckIfMethodIsApplicableAndPrepareIt(m, args))
    .ToArray();
  if (applicable.Length > 1)
  {
    // also returns an empty array if no overload has priority (ie. if there are multiple applicable methods)!
    return applicable
      .Where(m => applicable.All(n => m == n || MethodHasPriority(args, m, n)))
      .ToArray();
  }

  return applicable;
}
```

The issue is that the caller of `FindBestMethod` doesn't know if the resulting array is empty because no function was found, or if multiple applicable overloads were found.

We might be able to fix the issue like this, but it would require checking all calls to `FindBestMethod` to be sure:

```c#
private static MethodData[] FindBestMethod(IEnumerable<MethodData> methods, Expression[] args)
{
  // empty array if there are no applicable methods
  var applicable = methods
    .Where(m => CheckIfMethodIsApplicableAndPrepareIt(m, args))
    .ToArray();
  if (applicable.Length > 1)
  {
    var bestCandidates = applicable
      .Where(m => applicable.All(n => m == n || MethodHasPriority(args, m, n)))
      .ToArray();
    if (bestCandidates.length > 0)
      return bestCandidates; // should always have a length of 1 
  }

  return applicable;
}
```

src/DynamicExpresso.Core/Identifier.cs
davideicardi: Ok, get your point @metoule. As a first step we can add a new `ReplaceFunction` that always replace the function no matter if the same signature exist? So if the client wants to explicitly replace any function with the same name can calls this new function.
What do you think? @sagilio @metoule

src/DynamicExpresso.Core/Parsing/Parser.cs
davideicardi: I like it. Thank you for the explanation. @sagilio  Do you think you can review the code to follow this suggestion?

metoule:
I've created PR #231 to apply all suggestions.

src/DynamicExpresso.Core/Parsing/Parser.cs
sagilio: Thank you for the explanation, I already understand why to return 0 length array here. I think #231 has fixed this well, and will not have much performance impact when the overloads count is not very large.

PR #231, opened 2022-03-07
Fix registration of same method signature via SetFunction

### LINKED ISSUE #159: ParseException with ArgsIncompatibleWithDelegate message is throw when multiple functions are registered with the same signature

metoule:
This PR is the continuation of #229. 

- Fix #159 by ensuring the all applicable methods are returned when no overload is the best
- Add Unset methods to remove existing identifiers from the interpreter 
- SetFunction no longer allows multiple delegates with the same signature to be registered: only the last one is kept

### DIFF
--- a/src/DynamicExpresso.Core/Identifier.cs
+++ b/src/DynamicExpresso.Core/Identifier.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq.Expressions;
+using System.Reflection;
 
 namespace DynamicExpresso
 {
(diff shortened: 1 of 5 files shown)


metoule:
@sagilio this PR should fix all the issues you highlighted in #229

davideicardi (approved):
Nice!

sagilio:
@metoule @davideicardi Thank you! It looks works well. Could we make a new release for this?

davideicardi:
Just release version v2.12.0!

Issue #232, opened 2022-03-19
Interpreter Scope identifier support


halamah:
It would be very convenient to add a Scope support into Interpreter. Consider the following snippet:

```c#
record Person(string Name, string Gender, int Age);

var interpreter = new Interpreter();

// this is a proposed Scope feature
interpreter.UseScope<Person>();

var person = new Person("Ulrich", "Male", 30);

var lambda = interpreter.Parse("Gender == \"Male\" && Age >= 30"); 

Assert.True(lambda.Invoke(person));
```

So it looks like a `this` reference that is implicitly passed when calling class methods. 

Although I can pass a person object itself as a variables - this will make the expression more complex i.e. `"person.Gender == \"Male\" && person.Age >= 30"` what is less convenient and not concise.

davideicardi:
Yes, similar requirement came up also in the past, like #18  (back in 2014 ;-) ).
I agree that in some context it could be nice, on the other end it could add some complexity in the parser ... If you find some nice way to implement this feature I think it could be added.

halamah:
closing this one to track further in https://github.com/dynamicexpresso/DynamicExpresso/pull/233

PR #233, opened 2022-03-20
implicit `this` reference #232.

halamah:
Allows to implicitly access `this` identifier set either as a Variable or passed as a Parameter to have more concise expression.

In other word an emulation of `this` implicit argument passed to a method call.

```c#
const string Name = "John";
var context = new KeyValuePair<string, string>(nameof(Name), Name);
var interpreter = new Interpreter().SetVariable("this", context);

Assert.AreEqual(Name, interpreter.Eval("this.Value"));
Assert.AreEqual(Name, interpreter.Eval("Value"));
```

or

```c#
const string Name = "John";
var context = new KeyValuePair<string, string>(nameof(Name), Name);
var parameter = new Parameter("this", context.GetType());

Assert.AreEqual(Name, interpreter.Parse("this.Value", parameter).Invoke(context));
Assert.AreEqual(Name, interpreter.Parse("Value", parameter).Invoke(context));
```

https://github.com/dynamicexpresso/DynamicExpresso/issues/232

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,25 @@ Assert.AreEqual(30, myFunc.Invoke(23, 7));
 Assert.AreEqual(30, myFunc.Invoke(32, -2));
 ```
 
+### Special identifiers
+
+Either a variable or a parameter with name `this` can be referenced implicitly.
+
+```csharp
+class Customer { public string Name { get; set; } }
+
+var target = new Interpreter();
+
+// 'this' is treated as a special identifier and could by accessed implicitly 
+target.SetVariable("this", new Customer { Name = "John" });
+
+// explicit context reference via 'this' variable
+Assert.AreEqual("John", target.Eval("this.Name"));
+
+// 'this' variable is referenced implicitly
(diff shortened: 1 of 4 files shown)


metoule (commented):
LGTM!

metoule (approved):

davideicardi (approved):
LGTM! Thank you!

Can you just provide a little note about this feature inside the README documentation?
Maybe inside "Syntax and operators" section? I think it is important to highlight the fact that from now on `this` will be a special keyword ...

README.md
```diff
@@ -116,6 +116,25 @@ Assert.AreEqual(30, myFunc.Invoke(23, 7));
...
+### Special identifiers
+
+Either a variable or a parameter with name `this` can be referenced implicitly.
+
+```csharp
+class Customer { public string Name { get; set; } }
+
+var target = new Interpreter();
+
+// 'this' is treated as a special identifier and could by accessed implicitly
```
metoule, line 128: Typo: `by` should read `be`

halamah: Fixed

PR #239, opened 2022-06-10
Allow default values for generic parameters

israellot:
`public T MethodWithGenericParamAndDefault<T>(T a, T b = default)
			{
				return a;
			}`
The above method throws when parameter b is implicit. This PR searches for a previous specified parameter of the same type to construct the default expression for this case.


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -2111,7 +2111,26 @@ private static bool CheckIfMethodIsApplicableAndPrepareIt(MethodData method, Exp
...
 				if (x.HasDefaultValue)
-					return Expression.Constant(x.DefaultValue, x.ParameterType);
+				{
+					var parameterType = x.ParameterType;
+					if (x.ParameterType.ContainsGenericParameters)
+					{
+						//try finding an actual parameter for the generic
+						for(var i = 0; i < promotedArgs.Count; i++)
+						{
+							var actualArg = promotedArgs[i];
```
metoule, line 2121: Nitpick: this could be moved to line 2124 since it's not used in the `if` condition.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -2111,7 +2111,26 @@ private static bool CheckIfMethodIsApplicableAndPrepareIt(MethodData method, Exp
 			promotedArgs.AddRange(method.Parameters.Skip(promotedArgs.Count).Select<ParameterInfo, Expression>(x =>
 			{
 				if (x.HasDefaultValue)
-					return Expression.Constant(x.DefaultValue, x.ParameterType);
+				{
+					var parameterType = x.ParameterType;
+					if (x.ParameterType.ContainsGenericParameters)
```
metoule, line 2116: Can you check whether it shouldn't be `IsGenericParameter`? With `ContainsGenericParameters`, I'm wondering what would happen with a signature like this:

```
public List<T> MethodWithGenericParamAndDefault<T>(T a, List<T> b = default)
{
   return b;
}

var b = MethodWithGenericParamAndDefault(1);
Assert.Null(b);
```

metoule (changes requested):
It looks good, but I'd add the following test to be sure:

```
public List<T> MethodWithGenericParamAndDefault<T>(T a, List<T> b = default)
{
   return b;
}

var b = MethodWithGenericParamAndDefault(1);
Assert.Null(b);
```

src/DynamicExpresso.Core/Parsing/Parser.cs
israellot: Thanks! I've generalized the solution to work with multi level generic types like `List<List<T>>`

src/DynamicExpresso.Core/Parsing/Parser.cs
israellot: True

metoule (approved):
LGTM!

davideicardi (approved):
Thank you!

PR #242, opened 2022-06-22
Fix issue when parsing hexadecimal and binary literals

### LINKED ISSUE #241: Error when parsing hex number inside parenthesis
Hello,
Tool started to support hex (0x) and binary numbers (0b) starting with the latest release. However, I found a bug.

"(0x0)"  --> Parse error: Invalid integer literal '0x0)'
"(0x0 )" --> Successfully calculated

This bug causes trouble when functions are used. E.g. "MyFunction(0xABC)"

metoule:
The lexer took one more character than needed when parsing an hexadecimal or a binary literal:

```c#
4+0x124+5
```` 

Instead of `0x124`, the lexer considered `0x124+` to be the literal. Now, we backtrack one character to avoid this issue.

Fix #241


davideicardi (approved):
Thank you!

Issue #194, opened 2021-11-25
Support for List construction with values


waclaw66:
Would be possible to support syntax like this?....
`new List<int>() { 1, 2, 3 }`

metoule:
Relevant C# documentation: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#collection-initializers

> Collection initializers let you specify one or more element initializers when you initialize a collection type that implements 
`IEnumerable` and has `Add` with the appropriate signature as an instance method or an extension method. The element initializers can be a simple value, an expression, or an object initializer.

`new List<int>() { 1, 2, 3 }` 

is equivalent to:

```c#
List<int> list = new List<int>();
list.Add (1);
list.Add (2);
list.Add (3);
```

If the wanted `Add` method has multiple parameters, it's possible to specify them. For example, for a `Dictionary`:

```c#
var moreNumbers = new Dictionary<int, string>
{
    {19, "nineteen" },
    {23, "twenty-three" },
    {42, "forty-two" }
};
```

is equivalent to: 

```c#
var moreNumbers = new Dictionary<int, string>();
moreNumbers .Add(19, "nineteen");
moreNumbers .Add(23, "twenty-three");
moreNumbers .Add(42, "forty-two");
```

I also learned that you can use indexers!! 

```c#
var numbers = new Dictionary<int, string>
{
    [7] = "seven",
    [9] = "nine",
    [13] = "thirteen"
};
```

To be complete, we should probably also allow the creation of an `Array`:

`new int[] { 1, 2, 3 }`

israellot:
Supporting array initializer would be great as well
`new int[]{ 1, 2, 3}`

davideicardi:
Closed by #250 ! Thank you @holdenmai and @metoule !

PR #246, opened 2022-08-15
Added support of collection initializer syntax.

holdenmai:
Implementation of collection initializer syntax support for constructor #194   Copied a little bit of conflicting code from [pull 245](https://github.com/dynamicexpresso/DynamicExpresso/pull/245).

You can now write

```
target.Reference(typeof(System.Collections.Generic.List<>));
var intList = target.Eval<System.Collections.Generic.List<int>>("new List<int>(){1, 2, 3, 4, 5}");
Assert.AreEqual(5, intList.Count);
for (int i = 0; i < intList.Count; ++i)
{
    Assert.AreEqual(i + 1, intList[i]);
}
```

as well as implementations supporting multiple parameters in the Add method such as 

`new MyClassAdder(){{ 1, 2, 3, 4, 5},{\"6\" },7	}`
and
`new Dictionary<int, string>(){{1, \"1\"}, {2, \"2\"}, {3, \"3\"}, {4, \"4\"}, {5, \"5\"}}`

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1203,16 +1203,17 @@ namespace DynamicExpresso.Parsing
 
 		private Expression ParseMemberAndInitializerList(NewExpression newExpr, Type newType)
 		{
-			int originalPos = _token.pos;
+			var originalPos = _token.pos;
 			var bindingList = new List<MemberBinding>();
-			List<Expression> actions = new List<Expression>();
+			var actions = new List<Expression>();
 			ParameterExpression instance = null;
+			var allowCollectionInit = typeof(IEnumerable).IsAssignableFrom(newType);
 			while (true)
 			{
 				if (_token.id == TokenId.CloseCurlyBracket) break;
 				if (_token.id != TokenId.Identifier)
 				{
-					if (!typeof(IEnumerable).IsAssignableFrom(newType))
+					if (!allowCollectionInit)
 					{
 						throw CreateParseException(_token.pos, ErrorMessages.CollectionInitializationNotSupported, newType, typeof(IEnumerable));
 					}
@@ -1259,7 +1260,7 @@ namespace DynamicExpresso.Parsing
 				if (_token.id != TokenId.Comma) break;
 				NextToken();
 			}
-			if(instance != null)
+			if (instance != null)
 			{
 				actions.Add(instance);
 				return Expression.Block(new ParameterExpression[] { instance }, actions);
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Dynamic;
 using System.Globalization;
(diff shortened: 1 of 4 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1185,46 +1186,85 @@ private Expression ParseNew()
...
 			while (true)
 			{
 				if (_token.id == TokenId.CloseCurlyBracket) break;
-				ValidateToken(TokenId.Identifier, ErrorMessages.IdentifierExpected);
-
-				var propertyOrFieldName = _token.text;
-				var member = FindPropertyOrField(newType, propertyOrFieldName, false);
-				if (member == null)
-					throw CreateParseException(_token.pos, ErrorMessages.UnknownPropertyOrField, propertyOrFieldName, GetTypeName(newType));
+				if (_token.id != TokenId.Identifier)
```
metoule, line 1213: I understand why you perform the check, but that's a bit limiting, because an identifier can be used:

```c#
new List<string> { string.Empty } 
```

You can peek at the next token to check if it's a equal sign: if it's not, you can use `SetTextPos` to restart parsing at the wanted location and call `ParseExpressionSegment`.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1185,46 +1186,85 @@ private Expression ParseNew()
...
+				{
+					if (!typeof(IEnumerable).IsAssignableFrom(newType))
```
metoule, line 1215: Maybe extract a local variable, to avoid checking this at every iteration?

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1185,46 +1186,85 @@ private Expression ParseNew()
...
-			return bindings;
+			return initializedInstance;
 		}
 
-		private MemberBinding[] ParseMemberInitializerList(Type newType)
+		private Expression ParseMemberAndInitializerList(NewExpression newExpr, Type newType)
 		{
+			int originalPos = _token.pos;
 			var bindingList = new List<MemberBinding>();
+			List<Expression> actions = new List<Expression>();
```
metoule, line 1208: Please use `var`.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1185,46 +1186,85 @@ private Expression ParseNew()
...
 			ValidateToken(TokenId.CloseCurlyBracket, ErrorMessages.CloseCurlyBracketExpected);
 			NextToken();
-			return bindings;
+			return initializedInstance;
 		}
 
-		private MemberBinding[] ParseMemberInitializerList(Type newType)
+		private Expression ParseMemberAndInitializerList(NewExpression newExpr, Type newType)
 		{
+			int originalPos = _token.pos;
```
metoule, line 1206: Please use `var`.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1185,46 +1186,85 @@ private Expression ParseNew()
...
+					NextToken();
 
+					var value = ParseExpressionSegment();
+					bindingList.Add(Expression.Bind(member, value));
+				}
 				if (_token.id != TokenId.Comma) break;
 				NextToken();
 			}
-			return bindingList.ToArray();
+			if(instance != null)
```
metoule, line 1262: NIT: missing space before parenthesis.

metoule (commented):
This looks really good, thank you!! The only problem I can see is that you can't use any expression that starts with an identifier in the collection initializer list:

```c#
new List<string> { string.Empty }
```

is valid, but not accepted with your code.

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Updated in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Updated in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Updated in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Updated in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: I was having some trouble getting that to work when I was trying to account for the invalid case of 
`new MyClassAdder(){{StrProp = "SomeValue"}}`
but I will try again according to your review comments

holdenmai: Resolved in latest commit.  

Also accounted for the ability to perform collection of the .Add while simultaneously setting a variable/parameter value even if that matches a property/field name.

`var target = new Interpreter();
			target.Reference(typeof(MyClassAdder));
			Assert.DoesNotThrow(() => target.Eval<MyClassAdder>("new MyClassAdder(){{ 1, 2, 3, 4, 5},{StrProp = \"6\" },7}", new Parameter("StrProp", "0")));`

This is the equivalent of 
`string StrProp = "";
var mc = new MyClassAdder()
{
{1,2,3 },
{StrProp = "6" }
};
       
Assert.AreEqual(StrProp`

holdenmai: I was so focused on the improvements around the checking of the additional cases, I missed the original List<string> work.

davideicardi: Thank you @holdenmai . Also remember to update/rebase your branch so that we will not have conflicts during merge.

PR #250, opened 2022-08-20
Added support of collection initializer syntax.

holdenmai:
Implementation of collection initializer syntax support for constructor #194   Managed to somehow close the original PR on accident during rebase operation [pull 246](https://github.com/dynamicexpresso/DynamicExpresso/pull/246), but all comments have now been resolved.

You can now write

```
target.Reference(typeof(System.Collections.Generic.List<>));
var intList = target.Eval<System.Collections.Generic.List<int>>("new List<int>(){1, 2, 3, 4, 5}");
Assert.AreEqual(5, intList.Count);
for (int i = 0; i < intList.Count; ++i)
{
    Assert.AreEqual(i + 1, intList[i]);
}
or 
 target.Eval<System.Collections.Generic.List<string>>("new List<string>(){string.Empty}");
```

as well as implementations supporting multiple parameters in the Add method such as 

`new MyClassAdder(){{ 1, 2, 3, 4, 5},{\"6\" },7	}`
and
`new Dictionary<int, string>(){{1, \"1\"}, {2, \"2\"}, {3, \"3\"}, {4, \"4\"}, {5, \"5\"}}`

Reference to method parameters is also supported like below

`target.Eval<System.Collections.Generic.List<string>>("new List<string>(){StrProp = string.Empty}", new Parameter("StrProp", "0"))
target.Eval<System.Collections.Generic.List<string>>("new List<string>(){StrValue()}", new Parameter("StrValue", new Func<string>(() => "Func")))`

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1268,13 +1268,9 @@ namespace DynamicExpresso.Parsing
 				}
 				else if (_token.id != TokenId.Equal || _arguments.TryGetIdentifier(propertyOrFieldName, out _) || _arguments.TryGetParameters(propertyOrFieldName, out _))
 				{
-					SetTextPos(pos);
-					NextToken();
 					ParseCollectionInitalizer(newType, pos, bindingList, actions, instance, allowCollectionInit);
 					return;
 				}
-				SetTextPos(pos);
-				NextToken();
 			}
 			if (member == null)
 			{
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Dynamic;
 using System.Globalization;
@@ -1193,11 +1194,10 @@ namespace DynamicExpresso.Parsing
 			var constructor = applicableConstructors[0];
 			var newExpr = Expression.New((ConstructorInfo)constructor.MethodBase, constructor.PromotedParameters);
 
-			var memberBindings = new MemberBinding[0];
 			if (_token.id == TokenId.OpenCurlyBracket)
-				memberBindings = ParseObjectInitializer(newType);
+				return ParseWithObjectInitializer(newExpr, newType);
 
-			return Expression.MemberInit(newExpr, memberBindings);
+			return newExpr;
 		}
 
 		private Expression[] ParseArrayInitializerList()
@@ -1207,41 +1207,134 @@ namespace DynamicExpresso.Parsing
 				allowTrailingComma: true);
 		}
 
-		private MemberBinding[] ParseObjectInitializer(Type newType)
+		private Expression ParseWithObjectInitializer(NewExpression newExpr, Type newType)
 		{
 			ValidateToken(TokenId.OpenCurlyBracket, ErrorMessages.OpenCurlyBracketExpected);
 			NextToken();
-			var bindings = ParseMemberInitializerList(newType);
+			var initializedInstance = ParseMemberAndInitializerList(newExpr, newType);
 			ValidateToken(TokenId.CloseCurlyBracket, ErrorMessages.CloseCurlyBracketExpected);
 			NextToken();
-			return bindings;
+			return initializedInstance;
 		}
 
-		private MemberBinding[] ParseMemberInitializerList(Type newType)
+		private Expression ParseMemberAndInitializerList(NewExpression newExpr, Type newType)
 		{
+			var originalPos = _token.pos;
 			var bindingList = new List<MemberBinding>();
+			var actions = new List<Expression>();
+			var instance = Expression.Variable(newType);
+			var allowCollectionInit = typeof(IEnumerable).IsAssignableFrom(newType);
 			while (true)
 			{
 				if (_token.id == TokenId.CloseCurlyBracket) break;
-				ValidateToken(TokenId.Identifier, ErrorMessages.IdentifierExpected);
-
-				var propertyOrFieldName = _token.text;
-				var member = FindPropertyOrField(newType, propertyOrFieldName, false);
-				if (member == null)
-					throw CreateParseException(_token.pos, ErrorMessages.UnknownPropertyOrField, propertyOrFieldName, GetTypeName(newType));
-
-				NextToken();
-
-				ValidateToken(TokenId.Equal, ErrorMessages.EqualExpected);
-				NextToken();
-
-				var value = ParseExpressionSegment();
-				bindingList.Add(Expression.Bind(member, value));
-
+				if (_token.id != TokenId.Identifier)
+				{
+					ParseCollectionInitalizer(newType, originalPos, bindingList, actions, instance, allowCollectionInit);
+				}
+				else
+				{
+					ParsePossibleMemberBinding(newType, originalPos, bindingList, actions, instance, allowCollectionInit);
+				}
 				if (_token.id != TokenId.Comma) break;
 				NextToken();
 			}
-			return bindingList.ToArray();
+			if (bindingList.Count == 0)
+			{
+				actions.Insert(0, Expression.Assign(instance, newExpr));
+				actions.Add(instance);
+				return Expression.Block(new ParameterExpression[] { instance }, actions);
+			}
+			return Expression.MemberInit(newExpr, bindingList.ToArray());
+		}
+
+		private void ParsePossibleMemberBinding(Type newType, int originalPos, List<MemberBinding> bindingList, List<Expression> actions, ParameterExpression instance, bool allowCollectionInit)
+		{
+			ValidateToken(TokenId.Identifier, ErrorMessages.IdentifierExpected);
+
+			var propertyOrFieldName = _token.text;
+			var member = FindPropertyOrField(newType, propertyOrFieldName, false);
+			var pos = _token.pos;
+			if (allowCollectionInit)
+			{
+				NextToken();
+				//new T(){Prop = 1}
+				//new T(){variable = 2}
+				if (_token.id == TokenId.Equal && member != null)
+				{
+					if (actions.Count > 0)
+					{
+						throw CreateParseException(pos, ErrorMessages.InvalidInitializerMemberDeclarator);
+					}
+				}
+				else if (_token.id != TokenId.Equal || _arguments.TryGetIdentifier(propertyOrFieldName, out _) || _arguments.TryGetParameters(propertyOrFieldName, out _))
+				{
+					SetTextPos(pos);
+					NextToken();
+					ParseCollectionInitalizer(newType, pos, bindingList, actions, instance, allowCollectionInit);
+					return;
+				}
+				SetTextPos(pos);
+				NextToken();
+			}
+			if (member == null)
+			{
+				throw CreateParseException(pos, ErrorMessages.UnknownPropertyOrField, propertyOrFieldName, GetTypeName(newType));
+			}
+			NextToken();
+
+			ValidateToken(TokenId.Equal, ErrorMessages.EqualExpected);
+			NextToken();
+
+			var value = ParseExpressionSegment();
+			bindingList.Add(Expression.Bind(member, value));
+		}
+
+		private void ParseCollectionInitalizer(Type newType, int originalPos, List<MemberBinding> bindingList, List<Expression> actions, ParameterExpression instance, bool allowCollectionInit)
+		{
+			if (!allowCollectionInit)
+			{
+				throw CreateParseException(_token.pos, ErrorMessages.CollectionInitializationNotSupported, newType, typeof(IEnumerable));
+			}
+			if (bindingList.Count > 0)
+			{
+				throw CreateParseException(originalPos, ErrorMessages.InvalidInitializerMemberDeclarator);
+			}
+			if (_token.id == TokenId.OpenCurlyBracket)
+			{
+				var pos = _token.pos;
+				NextToken();
+
+				if (_token.id == TokenId.Identifier)
+				{
+					var identifierName = _token.text;
+					NextToken();
+					if (_token.id == TokenId.Equal && !_arguments.TryGetIdentifier(identifierName, out _) && !_arguments.TryGetParameters(identifierName, out _))
+					{
+						throw CreateParseException(_token.pos, ErrorMessages.InvalidInitializerMemberDeclarator);
+					}
+					else
+					{
+						SetTextPos(pos);
+						NextToken();
+					}
+				}
+				else
+				{
+					SetTextPos(pos);
+					ParseExpressionSegment();
+				}
+				actions.Add(ParseMethodInvocation(newType, instance, _token.pos, "Add", TokenId.OpenCurlyBracket, ErrorMessages.OpenCurlyBracketExpected, TokenId.CloseCurlyBracket, ErrorMessages.CloseCurlyBracketExpected));
+			}
+			else
+			{
+				var args = new[] { ParseExpressionSegment() };
+				var addMethod = ParseNormalMethodInvocation(newType, instance, _token.pos, "Add", args);
+				if (addMethod == null)
+				{
+					throw CreateParseException(_token.pos, ErrorMessages.UnableToFindAppropriateAddMethod, GetTypeName(newType));
+				}
+				actions.Add(addMethod);
+			}
 		}
 
 		private Expression ParseLambdaInvocation(LambdaExpression lambda, int errorPos)
@@ -1573,7 +1666,12 @@ namespace DynamicExpresso.Parsing
 
 		private Expression ParseMethodInvocation(Type type, Expression instance, int errorPos, string methodName)
 		{
-			var args = ParseArgumentList();
+			return ParseMethodInvocation(type, instance, errorPos, methodName, TokenId.OpenParen, ErrorMessages.OpenParenExpected, TokenId.CloseParen, ErrorMessages.CloseParenOrCommaExpected);
+
+		}
+		private Expression ParseMethodInvocation(Type type, Expression instance, int errorPos, string methodName, TokenId open, string openExpected, TokenId close, string closeExpected)
+		{
+			var args = ParseArgumentList(open, openExpected, close, closeExpected);
 
 			var methodInvocationExpression = ParseNormalMethodInvocation(type, instance, errorPos, methodName, args);
 			if (methodInvocationExpression == null && instance != null)
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1268,9 +1268,13 @@ namespace DynamicExpresso.Parsing
 				}
 				else if (_token.id != TokenId.Equal || _arguments.TryGetIdentifier(propertyOrFieldName, out _) || _arguments.TryGetParameters(propertyOrFieldName, out _))
 				{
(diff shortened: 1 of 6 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,126 @@ private Expression[] ParseArrayInitializerList()
...
+			return initializedInstance;
 		}
 
-		private MemberBinding[] ParseMemberInitializerList(Type newType)
+		private Expression ParseMemberAndInitializerList(NewExpression newExpr, Type newType)
 		{
+			var originalPos = _token.pos;
 			var bindingList = new List<MemberBinding>();
+			var actions = new List<Expression>();
+			ParameterExpression instance = Expression.Variable(newType);
```
metoule, line 1225: Please  always use var :)

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +179,84 @@ public void Array_multi_dimension_constructor()
...
+			}
+			Assert.AreEqual("6", l.StrProp);
+			Assert.AreEqual(7, l.IntField);
+		}
+
+
+		[Test]
+		public void Ctor_NewMyClassWithCross()
+		{
+			string StrProp = "";
```
metoule, line 214: Please always use var, and use `string.Empty`  insteaf of `""`

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +179,84 @@ public void Array_multi_dimension_constructor()
...
+			var mc = new MyClassAdder()
+			{
+				{1,2,3 },
+				{StrProp = "6" }
+			};
+			if (StrProp == "6")
+			{
+
+			}
```
metoule, line 223: Can be removed

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +179,84 @@ public void Array_multi_dimension_constructor()
...
+			var target = new Interpreter();
+			target.Reference(typeof(MyClassAdder));
+			Assert.DoesNotThrow(() => target.Eval<MyClassAdder>("new MyClassAdder(){{ 1, 2, 3, 4, 5},{StrProp = \"6\" },7}", new Parameter("StrProp", "0")));
+			Assert.DoesNotThrow(() => target.Eval<MyClassAdder>("new MyClassAdder(){{ 1, 2, 3, 4, 5},StrProp = \"6\",7}", new Parameter("StrProp", "0")));
```
metoule, line 227: If it's not expected to throw, can you instead test the expected results? In particular, that the expected value of StrProp is "6"?

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +179,84 @@ public void Array_multi_dimension_constructor()
...
+			var target = new Interpreter();
+			target.Reference(typeof(System.Collections.Generic.List<>));
+			var list = target.Eval<System.Collections.Generic.List<string>>("new List<string>(){string.Empty}");
+			Assert.AreEqual(1, list.Count);
+			for (int i = 0; i < list.Count; ++i)
+			{
+				Assert.AreSame(string.Empty, list[i]);
+			}
+			Assert.DoesNotThrow(() => target.Eval<System.Collections.Generic.List<string>>("new List<string>(){StrProp = string.Empty}", new Parameter("StrProp", "0")));
+			Assert.DoesNotThrow(() => target.Eval<System.Collections.Generic.List<string>>("new List<string>(){StrValue()}", new Parameter("StrValue", new Func<string>(() => "Func"))));
```
metoule, line 256: Please check the expected values.

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +179,84 @@ public void Array_multi_dimension_constructor()
...
+			Assert.DoesNotThrow(() => target.Eval<MyClassAdder>("new MyClassAdder(){{ 1, 2, 3, 4, 5},StrProp = \"6\",7}", new Parameter("StrProp", "0")));
+			Assert.DoesNotThrow(() => target.Eval<MyClassAdder>("new MyClassAdder(){{ 1, 2, 3, 4, 5},string.Empty, 7}"));
+			Assert.Throws<ParseException>(() => target.Eval<MyClassAdder>("new MyClassAdder(){{ 1, 2, 3, 4, 5},{StrProp = \"6\" },7	}"));
+		}
+		[Test]
+		public void Ctor_NewListWithItems()
+		{
+			var target = new Interpreter();
+			target.Reference(typeof(System.Collections.Generic.List<>));
+			var intList = target.Eval<System.Collections.Generic.List<int>>("new List<int>(){1, 2, 3, 4, 5}");
```
metoule, line 236: Can you add a test with a property access? For example, replace 5 with `int.MaxValue`, or a parameter? To ensure you can use any expression (I saw the next test uses a method call, but one more test won't hurt).

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,126 @@ private Expression[] ParseArrayInitializerList()
...
+						this._arguments.TryGetIdentifier(_token.text, out _)
+							||
+						FindPropertyOrField(newType, _token.text, false) == null)
+				{
+					ParseCollectionInitalizer(newExpr, newType, originalPos, bindingList, actions, instance, allowCollectionInit);
+					return;
+				}
+				else
+				{
+					throw CreateParseException(originalPos, ErrorMessages.InvalidInitializerMemberDeclarator);
```
metoule, line 1266: Can you please add a unit test for this exception?

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,126 @@ private Expression[] ParseArrayInitializerList()
...
-				if (_token.id != TokenId.Comma) break;
+			var value = ParseExpressionSegment();
+			bindingList.Add(Expression.Bind(member, value));
+		}
+
+		private void ParseCollectionInitalizer(NewExpression newExpr, Type newType, int originalPos, List<MemberBinding> bindingList, List<Expression> actions, ParameterExpression instance, bool allowCollectionInit)
+		{
+			if (!allowCollectionInit)
+			{
+				throw CreateParseException(_token.pos, ErrorMessages.CollectionInitializationNotSupported, newType, typeof(IEnumerable));
```
metoule, line 1303: Please add a unit test for this.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,126 @@ private Expression[] ParseArrayInitializerList()
...
+			}
+			
+			if (bindingList.Count > 0)
+			{
+				throw CreateParseException(originalPos, ErrorMessages.InvalidInitializerMemberDeclarator);
```
metoule, line 1308: Please add a unit test for this.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,126 @@ private Expression[] ParseArrayInitializerList()
...
+			{
+				var pos = _token.pos;
+				if (allowCollectionInit)
+				{
+					NextToken();
+					if (_token.id != TokenId.Equal || _arguments.TryGetIdentifier(propertyOrFieldName, out _) || _arguments.TryGetParameters(propertyOrFieldName, out _))
+					{
+						SetTextPos(pos);
+						NextToken();
+						actions.Add(ParseNormalMethodInvocation(newType, instance, _token.pos, "Add", new[] { ParseExpressionSegment() }));
```
metoule, line 1283: Should call `ParseCollectionInitalizer`

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,126 @@ private Expression[] ParseArrayInitializerList()
...
+			if (_token.id == TokenId.OpenCurlyBracket)
+			{
+				var pos = _token.pos;
 				NextToken();
+				if (_token.id == TokenId.Identifier)
+				{
+					if (!_arguments.TryGetIdentifier(_token.text, out var _) && !_arguments.TryGetParameters(_token.text, out var _))
+					{
+						throw CreateParseException(_token.pos, ErrorMessages.InvalidInitializerMemberDeclarator);
+					}
```
metoule, line 1319: I think this will prevent class constants from being used, such as `string.Empty`.

metoule (commented):
That looks really good! Apart from a few nitpicks, I'd recommend adding more tests, mostly to test the various possible exceptions. Also, I believe the current implementation can't parse `new List<string> { { string.Empty } }`.

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Incorporated in latest update

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Incorporated in latest update

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Added more positive and negative testing around generic list creation.

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Cleaned up the logic around transitioning over to collection init from member init process.

holdenmai:
I have added scenarios within the differing unit tests to cover each of the cases mentioned.  Is it preferred that they become standalone cases to better show what negative case they are trying to cover?

davideicardi:
Thank you @holdenmai 

> Is it preferred that they become standalone cases to better show what negative case they are trying to cover?
Yes, maybe having separate tests would be better... or try to simplify/explain better the existing ones.

holdenmai:
I have reorganized the unit tests to better show the different scenarios being covered.

davideicardi:
LGTM!

@metoule ? 

There is just 3 conversations still open that I'm not sure if they are now resolved ...

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Covered by Ctor_CannotUseCollectionInitDoesNotImplementIEnumerable test

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: covered in multiple tests under Ctor_InvalidInitializerMemberDeclarator test

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: The scenario of class constants has been checked in a couple different tests.
See Ctor_NewMyClass_ExpectedValues and Ctor_NewListWithItems for validation of different ways of referencing those values.

test/DynamicExpresso.UnitTest/ConstructorTest.cs
davideicardi: @holdenmai I think this is the last missing test. Or it is already done?

holdenmai: the Ctor_NewListWithItems has a lot of different tests calling out to different "constants", both direct and referencing out to int.MaxValue.

and Ctor_NewListWithString also contains references to using parameters.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,134 @@ private Expression[] ParseArrayInitializerList()
...
+				{
+					if (actions.Count > 0)
+					{
+						throw CreateParseException(pos, ErrorMessages.InvalidInitializerMemberDeclarator);
+					}
+				}
+				else if (_token.id != TokenId.Equal || _arguments.TryGetIdentifier(propertyOrFieldName, out _) || _arguments.TryGetParameters(propertyOrFieldName, out _))
+				{
+					SetTextPos(pos);
+					NextToken();
```
metoule, line 1272: Aren't these two operations NOP?

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1207,41 +1207,134 @@ private Expression[] ParseArrayInitializerList()
...
+					ParseCollectionInitalizer(newType, pos, bindingList, actions, instance, allowCollectionInit);
+					return;
+				}
+				SetTextPos(pos);
 				NextToken();
```
metoule, line 1277: Aren't these two operations NOP?

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +180,148 @@ public void Array_multi_dimension_constructor()
...
+		[Test]
+		public void Ctor_NewMyClass_CanStillUseMemberSyntax()
+		{
+			var target = new Interpreter();
+			target.Reference(typeof(MyClassAdder));
+			target.Reference(typeof(MyClass));
+			Assert.AreEqual(
+				new MyClassAdder() { StrProp = string.Empty, MyArr = new[] { 1, 2, 3, 4, 5 }, IntField = int.MinValue },
+				target.Eval<MyClassAdder>("new MyClassAdder() {StrProp = string.Empty, MyArr = new int[] {1, 2, 3, 4, 5}, IntField = int.MinValue }"));
+		}
```
metoule, line 236: Isn't this test case already handled by the previous test?

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +180,148 @@ public void Array_multi_dimension_constructor()
...
+			target.Reference(typeof(MyClassAdder));
+			target.Reference(typeof(MyClass));
+			var strProp = new Parameter("StrProp", typeof(string).MakeByRefType(), "0");
+			var args = new object[]
+			{
+				strProp.Value
+			};
+			Assert.AreEqual(
+				new MyClassAdder() { { 1, 2, 3, 4, 5 }, "6", 7 },
+				target.Parse("new MyClassAdder(){{ 1, 2, 3, 4, 5},{StrProp = \"6\" },7}", strProp).Invoke(args));
```
metoule, line 218: Nice test case, it took me a while to understand that this is not assigning the `StrProp` property, but instead assigning a value to the ref parameter `StrProp`!!

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +180,148 @@ public void Array_multi_dimension_constructor()
...
+			Ctor_NewListGeneric<object, object>("string.Empty", "int.MinValue");
+		}
+
+		[Test]
+		public void Ctor_NewListCantFindAddMethod()
+		{
+			var target = new Interpreter();
+			target.Reference(typeof(System.Collections.Generic.List<>));
+			Assert.Throws<ParseException>(() => target.Eval<System.Collections.Generic.List<int>>("string.Empty"));
+			Assert.Throws<ParseException>(() => target.Eval<System.Collections.Generic.List<string>>("int.MaxValue"));
```
metoule, line 274: I think you're missing the call to the List constructor here.

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +180,148 @@ public void Array_multi_dimension_constructor()
...
+			//Create a random list of values to test.
+			var actual = new System.Collections.Generic.List<TObject>();
+			foreach (var v in items)
+			{
+				actual.Add(target.Eval<TObject>(v.Trim('}', '{')));
+			}
+			for (var min = 0; min < actual.Count; ++min)
+			{
+				for (var count = Math.Min(min, 1); count <= actual.Count - min; ++count)
+				{
```
metoule, line 292: What's the purpose of the two loops? Wouldn't be enough to just concat all the provided items?

test/DynamicExpresso.UnitTest/ConstructorTest.cs
```diff
@@ -179,6 +180,148 @@ public void Array_multi_dimension_constructor()
...
+		[Test]
+		public void Ctor_NewListCantFindAddMethod()
+		{
+			var target = new Interpreter();
+			target.Reference(typeof(System.Collections.Generic.List<>));
+			Assert.Throws<ParseException>(() => target.Eval<System.Collections.Generic.List<int>>("string.Empty"));
+			Assert.Throws<ParseException>(() => target.Eval<System.Collections.Generic.List<string>>("int.MaxValue"));
+		}
+
+		public void Ctor_NewListGeneric<TList, TObject>(params string[] items)
```
metoule, line 277: TList and TObject are always the same, so you could use a single generic argument.

test/DynamicExpresso.UnitTest/ParametersTest.cs
```diff
@@ -332,6 +332,7 @@ public void When_lambda_is_invoked_input_parameters_must_follow_in_the_same_orde
 			Assert.AreEqual(4, lambda.Invoke(1, 5));
 		}
 
+
```
metoule, line 335: The changes to this file can be reverted.

metoule:
Sorry for the late review! The code looks good to me, the test cases as well, I still have a few questions, but nothing major.

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: If I take that out, it ends up causing the (now fixed) Ctor_NewListCantFindAddMethod test to fail with "Expression Expected" instead of cannot find appropriate Add method

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: In the case of  being a Prop1 = "Value" in our parsing, the token is at =.  Calling this sets the token back to the Identifier Prop1.  
If this is not called, the NextToken call a couple lines below ends up advancing us to "Value"

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Yes, missed in my splitting out of tests.

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Yeah, this test was actually the one that caused me to raise #252 as I technically have no way of validating that it actually did the assign.

metoule: No, I don't think that's needed for now, it's really an edge case.

holdenmai: completed.

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Correct

test/DynamicExpresso.UnitTest/ConstructorTest.cs
holdenmai: Updated

davideicardi (approved):

PR #253, opened 2022-08-22
Updating Lambda Invoke to support byref values.

### LINKED ISSUE #252: Lambda Invoke does not update arguments in the parameters array
Understandably, the Lambda class is essentially a wrapper around the parsed delegate due to the way we treat the parameter creation.

However, the way we currently implement the Invoke, the following fails.

```
		[Test]
		public void When_lambda_is_invoked_byref_parameters_are_updated_on_invoke()
		{
			var a = 1;
			var b = 2;
			var c = 3;
			var d = 4;
			var e = 5;

			var parameters = new[]{
							new Parameter(nameof(a), typeof(int).MakeByRefType()),
							new Parameter(nameof(b), typeof(int).MakeByRefType()),
							new Parameter(nameof(c), typeof(int).MakeByRefType()),
							new Parameter(nameof(d), typeof(int).MakeByRefType()),
							new Parameter(nameof(e), typeof(int).MakeByRefType()),
							};

			var args = new object[]
			{
				a,
				b,
				c,
				d,
				e
			};

			a += b *= c -= d /= e %= 3;

			var lambda = target.Parse("a = a + (b = b * (c = c - (d = d / (e = e % 3))))", parameters);

			Assert.AreEqual(a, lambda.Invoke(args));
			Assert.AreEqual(a, args[0]);
			Assert.AreEqual(b, args[1]);
			Assert.AreEqual(c, args[2]);
			Assert.AreEqual(d, args[3]);
			Assert.AreEqual(e, args[4]);
		}

```

If we had a reference to the actual delegate and performed a dynamic invoke on it with the args array, the Asserts would succeed.

holdenmai:
Implementation to resolve #252.

This needs to be completed prior to #250 since it depends on this logic to work in order to test some of the edge cases with combining variable/parameter setting in the Add methods.

Usage

```
var a = 1;

var parameters = new[]{
	new Parameter(nameof(a), typeof(int).MakeByRefType())
};

var lambda = target.Parse("2 + (a = a + 3)", parameters);
var args = new object[] { a };
var result = lambda.Invoke(args);
Assert.AreEqual(2 + (a + 3), result);
a += 3;
Assert.AreEqual(a, args[0]);
```

### DIFF
--- a/src/DynamicExpresso.Core/Lambda.cs
+++ b/src/DynamicExpresso.Core/Lambda.cs
@@ -83,23 +83,63 @@ namespace DynamicExpresso
 			var parameters = new List<Parameter>();
 			var declaredParameters = DeclaredParameters.ToArray();
 
-			if (args != null)
+			int[] actualArgOrdering = null;
+			object[] orderedArgs = args;
+			var argsAreReordered = false;
+			if (args != null && args.Length > 0)
 			{
 				if (declaredParameters.Length != args.Length)
 					throw new InvalidOperationException("Arguments count mismatch.");
 
+				actualArgOrdering = new int[args.Length];
+				var usedParametersIndex = new Dictionary<string, int>(_parserArguments.Settings.KeyComparer);
+				foreach (var v in UsedParameters)
+				{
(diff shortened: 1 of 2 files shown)


davideicardi:
Closing this for now. Write me if you think this is still relevant ...

Issue #256, opened 2022-09-18
Using Lambda expression inside a Lambda expression


mohamedabdelmonem-E:
When trying to call lambda expression form another lambda expression an exception is thrown (different exceptions when tweaking the code differently) below is an example of the code required to get evaluated:
```
using DynamicExpresso;
using System.Collections.Generic;
using System.Linq;

namespace TEST
{
    class Program
    {
        static void Main(string[] args)
        {
            ICollection<BounsMatrix> annualBouns = new List<BounsMatrix> {
                new() { Grade = 1, BounsFactor = 7 }, 
                new() { Grade = 2, BounsFactor = 5.5 }, 
                new() { Grade = 3, BounsFactor = 4 }, 
                new() { Grade = 4, BounsFactor = 3.5 }, 
                new() { Grade = 5, BounsFactor = 3 } 
            };

            ICollection<Employee> employees = new List<Employee> {
                new() { Id = "01", Name = "A", Grade = 5, Salary = 20000}, //bouns = 20000 * 7   = 60000
                new() { Id = "02", Name = "B", Grade = 5, Salary = 18000}, //bouns = 18000 * 7   = 54000
                new() { Id = "03", Name = "C", Grade = 4, Salary = 12000}, //bouns = 12000 * 5.5 = 42000
                new() { Id = "04", Name = "D", Grade = 4, Salary = 10000}, //bouns = 10000 * 5.5 = 35000
                new() { Id = "05", Name = "E", Grade = 3, Salary = 8500},  //bouns = 8500  * 4   = 34000
                new() { Id = "06", Name = "F", Grade = 3, Salary = 8000},  //bouns = 8000  * 4   = 32000
                new() { Id = "07", Name = "G", Grade = 2, Salary = 5000},  //bouns = 5000  * 3.5 = 27500
                new() { Id = "08", Name = "H", Grade = 2, Salary = 4750},  //bouns = 4750  * 3.5 = 26125
                new() { Id = "09", Name = "I", Grade = 1, Salary = 3500},  //bouns = 3500  * 3   = 24500
                new() { Id = "10", Name = "J", Grade = 1, Salary = 3250}   //bouns = 3250  * 3   = 22750
            };
            
            var interpreter = new Interpreter(InterpreterOptions.LambdaExpressions | InterpreterOptions.Default);
            interpreter.SetVariable(nameof(annualBouns), annualBouns);
            interpreter.SetVariable(nameof(employees), employees);

[...]

PR #257, opened 2022-09-19
Issues 212, 256: Adding support for internal lambdas to reference parameters from the parent lambda

### LINKED ISSUE #212: Parsing Lambda expression pre-evalutes it with given arguments

holdenmai:
Issue #212 was really the issue here.  

When replicating the code given in Issue #256 the last "x.Grade" was being immediately resolved as the value from the parameter was being inserted directly.  In the previous implementation of "InterpreterExpression", it was creating it as a new ParameterExpression.  
This code update captures the "parent" ParameterExpressions as identifiers.

This is essentially what happens with compiled code, so instead of us having to create a class, Expression actually allows us to reference the existing Expression as an Identifier.

Fix #212, #200, #256

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Dynamic;
 using System.Globalization;
 using System.Linq;
@@ -138,7 +139,7 @@ namespace DynamicExpresso.Parsing
 		{
 			// in case the expression is not a lambda, we have to restart parsing
 			var originalPos = _token.pos;
-
+			var isLambda = false;
 			try
 			{
 				var parameters = ParseLambdaParameterList();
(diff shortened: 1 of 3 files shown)


davideicardi:
LGTM! Thank you!

@metoule do you see problems?

davideicardi:
Do you think that this PR will solve #212, #200 and #256?

holdenmai:
Yes it will.  

This will also make it so that #213 can be closed without merging as the settings Identifier dictionary has become our "locals" dictionary.

I just thought of a couple more cases to verify that we are only able to access parameters defined in the enclosing lambda, not a parameter defined by an adjacent lambda.

Will get those included as soon as able.

holdenmai:
Test cases to cover duplicating parameter names in other lambdas have been added.

davideicardi:
Thank you @holdenmai . It looks like a very nice feature. There is just a failed test.

holdenmai:
Unit test has been fixed.  It actually had a bad lambda declaration that would not have compiled with real code.  I guess the only issue with that is if a consumer of this package had also done the same thing in the past.

metoule (approved):
That's even better than my attempt in #213 (and simpler :)), thanks a lot!!

Issue #93, opened 2018-10-30
Cannot access anonymous class in properties of dynamic object


NielsUll:
Accessing a property of an anonymous class fails when going through a dynamic object:

> Microsoft.CSharp.RuntimeBinder.RuntimeBinderException : 'object' does not contain a definition for 'Foo'

```
[Test]
public void Get_Property_of_a_nested_anonymous()
{
    dynamic dyn = new ExpandoObject();
    dyn.Sub = new { Foo = "bar"};
    var interpreter = new Interpreter()SetVariable("dyn", (object)dyn);
    Assert.AreEqual(dyn.Sub.Foo, interpreter.Eval("dyn.Sub.Foo"));
}
```

It works fine when the class is named: `dyn.Sub = new MyClass() { Foo = "bar"};` 
It also works if we introduce an intermediate variable:
```
    dynamic dyn = new ExpandoObject();
    dyn.Sub = new { Foo = "bar"};
    var interpreter = new Interpreter().SetVariable("dyn", dyn);
    interpreter.SetVariable("sub", interpreter.Eval("dyn.Sub"));
    Assert.AreEqual(dyn.Sub.Foo, interpreter.Eval("sub.Foo"));

```

Aslo discussed on stack overflow: https://stackoverflow.com/questions/53040955/how-can-i-use-dynamic-variables-with-anonymous-types

davideicardi:
Thanks @NielsUll Not sure when I can work on it, so any help is appreciated.

metoule:
In C#, anonymous types are internal types, which mean they're only visible to the assembly in which they're defined. In your scenario, the defining assembly (yours) is not the same as the DynamicExpresso assembly, which means the binder can't find its properties. There's a simple example here: https://stackoverflow.com/questions/2630370/c-sharp-dynamic-cannot-access-properties-from-anonymous-types-declared-in-anot . 

You can see the same behavior when trying to use a private class as `Sub`:

```csharp
[Test]
public void Get_Property_of_a_nested_internal_class()
{
	dynamic dyn = new ExpandoObject();
	dyn.Sub = new MyClass { Foo = "bar" };
	var interpreter = new Interpreter().SetVariable("dyn", (object)dyn);
	Assert.AreEqual(dyn.Sub.Foo, interpreter.Eval("dyn.Sub.Foo"));
}

private class MyClass
{
	public string Foo { get; set; }
}
```

If you change the class to be public instead, then the test pass.

[...]

PR #258, opened 2022-09-21
Adding fuller support for resolving anonymous type data on a dynamic object

holdenmai:
Issue #93 

The major change made with this is that rather than performing the binding, somewhat greedily, on typeof(object), we leverage the actual instance present in the CallSiteBinder args parameter to get the actual type as late as possible.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -1677,67 +1677,15 @@ namespace DynamicExpresso.Parsing
 		//    return Expression.Call(typeof(Enumerable), signature.Name, typeArgs, args);
 		//}
 
-		private class LateGetMemberCallSiteBinder : CallSiteBinder
-		{
-			private readonly string m_propertyOrFieldName;
-
-			public LateGetMemberCallSiteBinder(string propertyOrFieldName)
-			{
-				m_propertyOrFieldName = propertyOrFieldName;
-			}
-
-			public override Expression Bind(object[] args, ReadOnlyCollection<ParameterExpression> parameters, LabelTarget returnLabel)
-			{
-				var binder =
-					Microsoft.CSharp.RuntimeBinder.Binder.GetMember(
-					Microsoft.CSharp.RuntimeBinder.CSharpBinderFlags.None,
-					m_propertyOrFieldName,
-					RemoveArrayType(args[0].GetType()),
-					new[] { Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfo.Create(Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfoFlags.None, null) }
-				);
-				var e = binder.Bind(args, parameters, returnLabel);
-				return e;
-			}
-		}
-
-		private class LateInvokeMethodCallSiteBinder : CallSiteBinder
-		{
-			private readonly string _methodName;
-
-			public LateInvokeMethodCallSiteBinder(string methodName)
-			{
-				_methodName = methodName;
-			}
-
-			public override Expression Bind(object[] args, ReadOnlyCollection<ParameterExpression> parameters, LabelTarget returnLabel)
-			{
-				var binderM = Microsoft.CSharp.RuntimeBinder.Binder.InvokeMember(
-					Microsoft.CSharp.RuntimeBinder.CSharpBinderFlags.None,
-					_methodName,
-					null,
-					RemoveArrayType(args[0].GetType()),
-					parameters.Select(x => Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfo.Create(Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfoFlags.None, null))
-				);
-				return binderM.Bind(args, parameters, returnLabel);
-			}
-		}
-
-		private class LateInvokeIndexCallSiteBinder : CallSiteBinder
-		{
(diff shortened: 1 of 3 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1676,43 +1677,38 @@ private Expression ParseNormalMethodInvocation(Type type, Expression instance, i
...
 		{
-			var binder = Microsoft.CSharp.RuntimeBinder.Binder.GetMember(
-				Microsoft.CSharp.RuntimeBinder.CSharpBinderFlags.None,
-				propertyOrFieldName,
-				type,
-				new[] { Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfo.Create(Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfoFlags.None, null) }
-			);
+			if (t == null || t.IsArray)
+			{
+				return null;
```
metoule, line 1690: If it works with null, it doesn't really matter, but maybe the binder requires the array's element type?

```c#
return t.GetElementType();
```

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -3340,5 +3336,73 @@ private static T[] RemoveLast<T>(T[] array)
 				return result;
 			}
 		}
+
+		/// <summary>
+		/// Binds to a member access of an instance as late as possible.  This allows the use of anonymous types on dynamic values.
+		/// </summary>
+		private class LateGetMemberCallSiteBinder : CallSiteBinder
+		{
+			private readonly string m_propertyOrFieldName;
```
metoule, line 3345: To be consistent with naming conventions, please use `_propertyOrFieldName`

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -3340,5 +3336,73 @@ private static T[] RemoveLast<T>(T[] array)
...
+			{
+				var binder =
+					Microsoft.CSharp.RuntimeBinder.Binder.GetMember(
+					Microsoft.CSharp.RuntimeBinder.CSharpBinderFlags.None,
+					m_propertyOrFieldName,
+					RemoveArrayType(args[0]?.GetType()),
+					new[] { Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfo.Create(Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfoFlags.None, null) }
+				);
+				var e = binder.Bind(args, parameters, returnLabel);
+				return e;
```
metoule, line 3362: `return  binder.Bind(args, parameters, returnLabel);`

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -3340,5 +3336,73 @@ private static T[] RemoveLast<T>(T[] array)
...
+
+			public LateGetMemberCallSiteBinder(string propertyOrFieldName)
+			{
+				m_propertyOrFieldName = propertyOrFieldName;
+			}
+
+			public override Expression Bind(object[] args, ReadOnlyCollection<ParameterExpression> parameters, LabelTarget returnLabel)
+			{
+				var binder =
+					Microsoft.CSharp.RuntimeBinder.Binder.GetMember(
```
metoule, line 3355: [NIT] `var binder = Microsoft.CSharp.RuntimeBinder.Binder.GetMember(` on the same line

metoule (commented):
I have a few minor nitpicking comments, but otherwise it looks good to me, thanks a lot!

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: It seems like it works just fine whether we return null or the element type.  This might be a detail that will require a very specific edge case for someone to encounter that isn't already covered by the unit tests to figure out which one is needed.

test/DynamicExpresso.UnitTest/DynamicTest.cs
```diff
@@ -20,7 +20,20 @@ public void Get_Property_of_an_ExpandoObject()
 			var interpreter = new Interpreter()
 				.SetVariable("dyn", (object)dyn);
 
-			Assert.AreEqual(dyn.Foo, interpreter.Eval("dyn.Foo"));
+			var l = interpreter.Parse("dyn.Foo");
+			l.Invoke();
+			l.Invoke();
+
+			//Assert.AreEqual(dyn.Foo, interpreter.Eval("dyn.Foo"));
```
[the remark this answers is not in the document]
holdenmai: Reverted/resolved in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Resolved in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Resolved in latest commit

src/DynamicExpresso.Core/Parsing/Parser.cs
holdenmai: Resolved in latest commit

davideicardi (approved):
LGTM! Very nice improvements!

Thank you @holdenmai !

PR #260, opened 2022-10-03
Consider curly brackets to decide when the body of a lambda stops

### LINKED ISSUE #259: Select Linq not working with more than one parameter
Hello Team, 

I want to use following expression but it is throwing error "UnknownIdentifierException 'visualCount' ". Following expression is fine fine when select linq has only single property i.e. pageName or visualCount.

```
var options = InterpreterOptions.Default | InterpreterOptions.LambdaExpressions; 
var interpreter = new Interpreter(options).SetVariable("courseList", courseList);

string expression = @"courseList.Select(x =>  new { pageName = x.PageName , visualCount = 5 })";

Lambda parsedExpression = interpreter.Parse(expression);
var result = parsedExpression.Invoke();

```

Note: I have tried every property, each time it worked for single property but not for multiple properties.

Please resolve this issue as I am currently blocked.

metoule:
Fix #259

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -149,6 +149,8 @@ namespace DynamicExpresso.Parsing
 
 				var startExpr = _parsePosition;
 
+				// decide where the lambda expression body ends:
+				// it's the first unmatched parenthesis, curly bracket, of comma
 				var parenCount = 0;
 				var inLambdaBody = true;
 				while (inLambdaBody)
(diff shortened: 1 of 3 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -149,16 +149,18 @@ private Expression ParseLambdaExpression()
 
 				var startExpr = _parsePosition;
 
+				// decide where the lambda expression body ends:
+				// it's the first unmatched parenthesis, curly bracket, of comma
```
holdenmai, line 153: Would this "succeed" on the following code?

`courseList.Select(x => new PageType() { PageName = x.PageName, VisualCount = 5 ))`

metoule: Yes, the matching will "work", but this algorithm is only here to detect where the body of a lambda expression ends, so that we know which string to parse. That string is then parsed independently.

Here the lambda expression body will be considered to be `x => new PageType() { PageName = x.PageName, VisualCount = 5 )`, but that's not a valid expression, and therefore the parsing will fail.

holdenmai: yes

metoule: Yes, sorry!

Issue #270, opened 2022-12-13
Add support of .net 7.0


zatuliveter:
I tried to run unit tests on the .net 7.0, everything is working except of linq stuff.
The following unit tests are failing:
- EnumerableTest.Invoke_enumerable_extensions
- GithubIssues.Lambda_Issue_256
- LambdaExpressionTest.Sum
- LambdaExpressionTest.Sum_string_length

`Error Message: 
System.ArgumentException : GenericArguments[1], 'System.Object', on 'TResult Average[TSource,TAccumulator,TResult](System.Collections.Generic.IEnumerable1[TSource])' violates the constraint of type 'TAccumulator'. System.Security.VerificationException : Method System.Linq.Enumerable.Average: type argument 'System.Object' violates the constraint of type parameter 'TAccumulator'.`

Linq also fails if you try to use nuget on .net 7.0 project (.net 6.0 are OK).

davideicardi:
I suspect that the reason is because these packages are not officially intended to be used directly... but I think that in some way we can adapt the code, in the worst case using a conditional compilation. As usual any help is appreciated.

metoule:
It's actually not related to .NET7, you can reproduce it with this code:

```c#
[Test]
public void Method_with_generic_constraints()
{
  var target = new Interpreter();

  var x = new MyTestService();
  target.SetVariable("x", x);

  Assert.AreEqual("works", target.Eval("x.GenericMethodWithConstraint(\"works\")"));
  Assert.Throws<NoApplicableMethodException>(() => target.Eval("x.GenericMethodWithConstraint(5)"), "This shouldn't throw a System.ArgumentException \"Violates the constraint of type 'T'\"");
}

private class MyTestService
{
  public T GenericMethodWithConstraint<T>(T input) where T : class
  {
    return input;
  }
}
```

You can't call `MyTestService.GenericMethodWithConstraint` with an integer, as it violates the generic constraints. DynamicExpresso tries to use the method, which triggers a generic constraint violation exception.

PR #271, opened 2022-12-20
.NET 7 build updates #270

trentbagshaw:
Sum and Average have new type parameter options that cause issues at runtime. Simple bypass operation that should only affect .NET 7 compilation. Affects #270

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -2314,7 +2314,12 @@ namespace DynamicExpresso.Parsing
 			if (method.MethodBase != null && method.MethodBase.IsGenericMethodDefinition &&
 			    method.MethodBase is MethodInfo)
 			{
-				var genericMethod = MakeGenericMethod(method);
+				MethodInfo genericMethod;
+#if NET7_0_OR_GREATER
+				genericMethod = TryMakeGenericMethod(method);
+#else
+				genericMethod = MakeGenericMethod(method);
+#endif
 				if (genericMethod == null)
 					return false;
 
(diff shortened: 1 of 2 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -2314,7 +2314,12 @@ private static bool CheckIfMethodIsApplicableAndPrepareIt(MethodData method, Exp
 			if (method.MethodBase != null && method.MethodBase.IsGenericMethodDefinition &&
 			    method.MethodBase is MethodInfo)
 			{
-				var genericMethod = MakeGenericMethod(method);
+				MethodInfo genericMethod;
+#if NET7_0_OR_GREATER
+				genericMethod = TryMakeGenericMethod(method);
+#else
+				genericMethod = MakeGenericMethod(method);
+#endif
```
metoule, line 2322: You can simply replace `MakeGenericMethod` with `TryMakeGenericMethod`, and avoid a compilation target check.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -2364,12 +2369,41 @@ private static MethodInfo MakeGenericMethod(MethodData method)
...
 		private static Dictionary<string, Type> ExtractActualGenericArguments(
 			Type[] methodGenericParameters,
 			Type[] methodActualParameters)
 		{
 			var extractedGenericTypes = new Dictionary<string, Type>();
 
+#if NET7_0_OR_GREATER
+			if (methodGenericParameters.Length != methodActualParameters.Length)
+				return extractedGenericTypes;
+#endif
```
metoule, line 2405: I'm surprised this situation can occur. In which situation did you see it?

metoule:
Closed in favor of #274

PR #289, opened 2023-06-10
No longer move to the next token when parsing an `is` or `as` expression

### LINKED ISSUE #287: Parsing exception while using 'as' operator
I'm trying to use following expression:

`string.IsNullOrEmpty(variable as string)`

but it results with exception: DynamicExpresso.Exceptions.ParseException: ')' or ',' expected (at index 40).

Adding extra parenthesis as message suggests solves the issue, but the expression is obviously wrong then
`string.IsNullOrEmpty(variable as string))`

Seems there is something wrong with parser.

metoule:
After the changes introduced in #178, the parser no longer needs to move to the token after the known type.

Fixes #287

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -481,8 +481,6 @@ namespace DynamicExpresso.Parsing
 					left = Expression.TypeAs(left, knownType);
 				else
 					throw CreateParseException(_token.pos, ErrorMessages.SyntaxError);
-
-				NextToken();
 			}
 
 			return left;
@@ -1443,6 +1441,22 @@ namespace DynamicExpresso.Parsing
 			var originalPos = _token.pos;
 			_arguments.TryGetKnownType(name, out type);
 
+			type = ParseKnownGenericType(name, type);
+			type = ParseTypeModifiers(type);
+
+			if (type == null)
+			{
+				// type name couldn't be parsed: restore position
+				SetTextPos(originalPos);
+				NextToken();
+				return false;
+			}
+
+			return true;
+		}
+
+		private Type ParseKnownGenericType(string name, Type type)
+		{
 			NextToken();
 			if (_token.id == TokenId.LessThan)
 			{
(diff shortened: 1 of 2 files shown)


src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -1443,6 +1441,22 @@ private bool TryParseKnownType(string name, out Type type)
 			var originalPos = _token.pos;
 			_arguments.TryGetKnownType(name, out type);
 
+			type = ParseKnownGenericType(name, type);
```
metoule, line 1444: This change is purely cosmetic: I've introduced a new method `ParseKnownGenericType` to make the code a bit easier to understand.

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -3346,6 +3350,9 @@ private void NextToken()
 					}
 					if (_parsePosition == _expressionTextLength)
 					{
+						if (_token.id == TokenId.End)
+							throw new InvalidOperationException("NextToken called when already at the end of the expression");
```
metoule, line 3354: I've added an exception if the parser tries to move past the end of the expression. The error in #287 would have been detected with this.

test/DynamicExpresso.UnitTest/GithubIssues.cs
```diff
@@ -510,7 +510,7 @@ public void GitHub_Issue_221_Case_insensitivity()
 
 			// case insensivity outside lambda expressions
 			Assert.IsFalse(interpreter.Eval<bool>("dateinthepast > now()")); // identifier
-			Assert.IsTrue(interpreter.Eval<bool>("dateinthepast is datetimeoffset)")); // known type
+			Assert.IsTrue(interpreter.Eval<bool>("dateinthepast is datetimeoffset")); // known type
```
metoule, line 513: One the test had an invalid parenthesis to workaround the parser's bug 😅

src/DynamicExpresso.Core/Parsing/Parser.cs
```diff
@@ -481,8 +481,6 @@ private Expression ParseTypeTesting()
 					left = Expression.TypeAs(left, knownType);
 				else
 					throw CreateParseException(_token.pos, ErrorMessages.SyntaxError);
-
-				NextToken();
```
metoule, line 485: That's the actual code fix

test/DynamicExpresso.UnitTest/GithubIssues.cs
davideicardi: Didn't remember 😅

davideicardi (approved):
Great PR as usual! Thank you!

Maybe you can just add as test the case 'string.IsNullOrEmpty(variable as string))'?
The one mentioned in the bug...

metoule:
Good idea, I've added the test for the case mentioned in the bug report

davideicardi (approved):

PR #293, opened 2023-07-14
Add support for Action<> lambdas

### LINKED ISSUE #292: Can lambda expressions only be used as conditional judgments and return bool?
To implement the mocked foreach, I use an extension method:
```
public static class Extend
{
     public static void All<T>(this IEnumerable<T> source, Action<T> action)
     {
         foreach (var item in source)
             action(item);
     }
}
```

I want to change the money field of each npc in `List<Npc> NearNpcs` to 10, so I wrote a Lambda string:

`"NearNpcs.All(n=>n.money=10)"`

(The `InterpreterOptions.LambdaExpressions` setting is already turned on, and the `Extend` class and associated required variables are registered.)

It reported the error:

> DynamicExpresso.Exceptions.ParseException: A value of type 'Int32' cannot be converted to type 'Boolean' (at index 0).

Does Lambda parsing currently only support conditional parsing that can return bool?

metoule:
Fix #292

### DIFF
--- a/src/DynamicExpresso.Core/Lambda.cs
+++ b/src/DynamicExpresso.Core/Lambda.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
+using DynamicExpresso.Reflection;
 
 namespace DynamicExpresso
 {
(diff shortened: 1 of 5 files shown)


davideicardi (approved):
Great! :rocket: Thank you!

Issue #295, opened 2023-09-14
dynamic/ExpandoObject string member can't be passed to a function that accepts a string


AndrewKoransky:
If you add a variable to the interpreter that is a dynamic/ExpandoObject with a string member, then attempt to use that string member in a function, you get an exception:
`DynamicExpresso.Exceptions.ParseException : Argument list incompatible with delegate expression (at index 0).`

Here is a test that demonstrates the behavior.

```cs
var evaluator = new Interpreter();

// create path helper functions in expressions...
Func<string, string, string> pathCombine = (x, y) => System.IO.Path.Combine(x, y);
evaluator.SetFunction("PathCombine", pathCombine);

// add a GlobalSettings dynamic object...
dynamic globalSettings = new ExpandoObject();
globalSettings.MyTestPath = "C:\\delme\\";
evaluator.SetVariable("GlobalSettings", globalSettings);

// Here is a workaround that succeeds - you have to cast GlobalSettings.MyTestPath to string:
string works = (string)evaluator.Eval("PathCombine((string)GlobalSettings.MyTestPath,\"test.txt\")");
Assert.That(works, Is.EqualTo("C:\\delme\\test.txt"));

// I think this should work, but doesn't:
string doesntWork = (string)evaluator.Eval("PathCombine(GlobalSettings.MyTestPath,\"test.txt\")");
Assert.That(doesntWork, Is.EqualTo("C:\\delme\\test.txt"));
```

Because PathCombine takes a string, and GlobalSettings.MyTestPath is a string, I think this should work?

metoule:
Since #296 has been reverted in #333, let's reopen this issue.

metoule:
I had a deeper look at the issue. We don't support method calls with dynamic arguments, at all, even for methods not set via `SetFunction`. For example:

```c#
dynamic dyn = new ExpandoObject();
dyn.Foo = "test";

var interpreter = new Interpreter()
	.SetVariable("dyn", (object)dyn);

Assert.That(interpreter.Eval("string.IsNullOrEmpty(dyn.Foo)"), Is.EqualTo(string.IsNullOrEmpty(dyn.Foo)));
```

throws `DynamicExpresso.Exceptions.NoApplicableMethodException : No applicable method 'IsNullOrEmpty' exists in type 'String' (at index 7).`

While it looks similar to the reported bug, the fix will be different because the binder to use is slightly different. I should be able to fix the two issues.

PR #296, opened 2023-09-14
Fix Issue #295

Pentiva:
This fixes issue #295 and does not fail any other tests, so I think this is probably the correct solution. 
Whether this should be extended/limited in some way based on InterpreterOptions.LateBindObject, I'm not sure.

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -2459,7 +2459,7 @@ namespace DynamicExpresso.Parsing
 					return Expression.Convert(expr, genericType);
 			}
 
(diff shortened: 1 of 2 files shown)


davideicardi:
Thank you @Pentiva!
I see tests are failing now because `Path` is not defined. I suggest to use just a static string.

Regarding the solution I don't see problem, but not sure if there is some edge case that I'm not aware. What do you think @metoule ?

Pentiva:
Sorry about that, I was using the online IDE github.dev for the first time and didn't notice it didn't add the usings automatically. 
I think I'll stick to using my local computer for now.

davideicardi (approved):
LGTM! Thank you!

PR #306, opened 2024-04-09
Fix Issue #305

### LINKED ISSUE #305: Indexer ambiguity when inherited class replaces parent's indexer with new keyword
Hello! 

First of all, thanks to all of the contributors for this project.

This is a very rare case, but it exists in our project. The issue happens when descendant class overrides indexer with `new` keyword and sets a different return type. In this case, `Parser` [finds two applicable methods](https://github.com/dynamicexpresso/DynamicExpresso/blob/2696232a8dbe1c52371d0a59602e6c9e60ddd781/src/DynamicExpresso.Core/Parsing/Parser.cs#L2145), calls `MethodHasPriority`, and then [fails here](https://github.com/dynamicexpresso/DynamicExpresso/blob/2696232a8dbe1c52371d0a59602e6c9e60ddd781/src/DynamicExpresso.Core/Parsing/Parser.cs#L2714) because `MethodBase` is `null` for `IndexerData`.

Unit test code to reproduce:

``` c#
public class A
{
	public string this[int index] => "some string";
}

public class B : A
{
	public new int this[int index] => 25;
}

[Test]
public void Test()
{
	var b = new B();

	var lambda = interpreter.Parse("this[0]", new Parameter("this", b));
	var res = lambda.Invoke(b);
	Assert.AreEqual(25, res);
}
```
I investigated the code, and, if this is ok, I would like to create a pull request with suggested solution for this issue.

denbell5:
As described in #305 , the bug is related to overridden indexer in inherited class. So, to fix it, I added an additional comparison to select indexer with higher priority - indexer that is declared in the descendant class.

Fix #305

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -2711,7 +2711,10 @@ namespace DynamicExpresso.Parsing
 			if (better)
 				return true;
 
-			if (!method.MethodBase.IsGenericMethod && otherMethod.MethodBase.IsGenericMethod)
+			if (method.MethodBase != null &&
+				otherMethod.MethodBase != null &&
+				!method.MethodBase.IsGenericMethod &&
+				otherMethod.MethodBase.IsGenericMethod)
 				return true;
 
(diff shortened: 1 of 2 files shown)


metoule (commented):
Thank you!

PR #315, opened 2024-08-17
No longer try to parse a generic type if there's no possible candidates

### LINKED ISSUE #314: Misleading Parser error on '<' with missing Parameter
When a '<' follows after an unknown identifier a `'>' expected` Error is generated not `Unknown identifier`

Reproduce in RoslynPad:
```
#r "nuget: DynamicExpresso.Core, 2.16.1"

using DynamicExpresso;
var i = new Interpreter();
//var x= i.Eval("a < 1 || b > 1 ", new [] {new Parameter("a", -1.0m)}); // -> Unknown identifier 'b' (at index 9). [as expected]
var x= i.Eval("a < 1 || b < 1 ", new [] {new Parameter("a", -1.0m)}); // -> '>' expected (at index 13). [misleading]

```

metoule:
With the current algorithm that parses known types, we first try to parse the type argument list before checking if there's a matching generic type definition in the known types. This leads to an improper error message when trying to perform a less than comparison on a unknown identifier, because we try to parse the comparison as the type argument list 😅 

```c#
var i = new Interpreter();
var x= i.Eval("b < 1"); // ParseException: '>' expected (at index 4). [misleading]
```

Fix #314

### DIFF
--- a/src/DynamicExpresso.Core/ParserArguments.cs
+++ b/src/DynamicExpresso.Core/ParserArguments.cs
@@ -77,7 +77,7 @@ namespace DynamicExpresso
 		}
 
 		/// <summary>
-		/// Returns true if the known types contain a generic definition type with the given name + any arity (e.g. name`1).
+		/// Returns true if the known types contain a generic type definition with the given name + any arity (e.g. name`1).
 		/// </summary>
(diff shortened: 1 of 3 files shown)


davideicardi (approved):
:pray:

PR #317, opened 2024-08-18
Remove unneeded NETSTANDARD1_6 compiler directives and unneeded usings

metoule:
Since we only compile for netstandard2.0, the compiler directive is no longer needed.


davideicardi (approved):
Thank you!

PR #318, opened 2024-08-18
🚚 Split Parser.cs

metoule:
The `Parser.cs` class is huge, which prevents code reuse: for example, we could imagine reuse the resolution logic in the late binders for dynamic objects.

In any case, having smaller files should make the code easier to understand. Let me know what you think :)

The PR has been split into several smaller commits to make it easier to review (and to validate each step).


davideicardi (approved):
Completely agree! The Parser class is very old and was never refactored. Thank you!

PR #319, opened 2024-09-10
Add german translations

cpiber:
This PR adds German translations

It seems like it's not possible to create translations for an external assembly using ResourceManager, and we need the errors in German.
I noticed this localization was added in #36, if @davideicardi or @jasonhumberstone know how I could provide translations from my project I would also be open to that.

If you ever need updates for the translations feel free to ping me.
I based the translations based on my understanding (mother tongue) and Microsoft's documentation translation.

### DIFF
--- a/src/DynamicExpresso.Core/Resources/ErrorMessages.de.resx
+++ b/src/DynamicExpresso.Core/Resources/ErrorMessages.de.resx
@@ -1,0 +1,180 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<root>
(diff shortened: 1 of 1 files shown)


davideicardi:
Thank you!
I don't see problems adding translations here.

Can you just check what happen if for some reason one translation is missing? I don't remember the default behavior ...

cpiber:
The default behavior is to fall back to the fallback resource (no language identifier), in this case English

davideicardi (approved):

cpiber:
Any plans for a new version with the localization included?

davideicardi:
Just released 2.17.1 ! :rocket:

PR #323, opened 2024-11-01
DetectIdentifier obtain complete variable name

alesanchez-windifferent:
Add logic to enable detection of complete variable name (ex. contact.firstname)

### DIFF
--- a/src/DynamicExpresso.Core/Detector.cs
+++ b/src/DynamicExpresso.Core/Detector.cs
@@ -11,21 +11,33 @@ namespace DynamicExpresso
 	{
 		private readonly ParserSettings _settings;
 
-		private static readonly Regex IdentifiersDetectionRegex = new Regex(@"(?<id>@?[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}_]*)", RegexOptions.Compiled);
+		private static readonly Regex RootIdentifierDetectionRegex =
+			new Regex(@"(?<id>@?[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}_]*)", RegexOptions.Compiled);
 
-		private static readonly string Id = IdentifiersDetectionRegex.ToString();
+		private static readonly Regex ChildIdentifierDetectionRegex = new Regex(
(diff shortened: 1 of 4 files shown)


src/DynamicExpresso.Core/Detector.cs
```diff
@@ -11,21 +11,33 @@ internal class Detector
...
+		private static readonly Regex CharDetectionRegex =
+			new Regex(@"(?<!\\)?'.{1,2}?(?<!\\)'", RegexOptions.Compiled);
 
 		public Detector(ParserSettings settings)
 		{
 			_settings = settings;
 		}
 
-		public IdentifiersInfo DetectIdentifiers(string expression)
+		public IdentifiersInfo DetectIdentifiers(string expression, bool includeChildName = false)
```
davideicardi, line 40: Instead of change the signature of the current `DetectIdentifiers`, can you create a new overload that takes an enum flag with that new option. Something like:

```
public IdentifiersInfo DetectIdentifiers(string expression)
{
  return this.DetectIdentifiers(expression, DetectOptions.None);
}

public IdentifiersInfo DetectIdentifiers(string expression, DetectOptions detectOptions)
{
  ...
}
```

In this way we do not break binary compatibility and it is more extensible for the future.
What do you think?

davideicardi (changes requested):

davideicardi:
Thank you for the PR. I only proposed a change in the signature.

src/DynamicExpresso.Core/Detector.cs
alesanchez-windifferent: Thx for the comment, applied.
Created the Enum and the method overload

davideicardi (approved):

metoule:
@alesanchez-windifferent can you please explain what you were trying to accomplish with this PR? The implementation seems wrong, and led to #351. For example, with your implementation:

```C#
target.SetVariable("test", "Toto");

var detectedIdentifiers = target.DetectIdentifiers("test.Length", DetectorOptions.IncludeChildren);
Assert.That(detectedIdentifiers.UnknownIdentifiers, Is.Empty);
```

Failure:
```
detectedIdentifiers.UnknownIdentifiers:
  Expected: <empty>
  But was:  < "test.Length" >
``` 

`test` is a valid variable, and `Length` is a valid property of `string`, so the Unknown identifiers should be empty. 

With `DetectorOptions.IncludeChildren`, you check whether the entire string `test.Length` is an identifier, or a lambda parameter, or a known type, but I don't see in which case it can ever true?

davideicardi:
@metoule I thought that with the PR the idea was to also check for any missing sub-properties or sub-functions. But I realize only now that the implementation is not correct. My mistake for not checking more carefully. Sorry.

Can we ollback the code, leaving DetectorOptions for the future, and marking IncludeChildren option as obsolete?

@alesanchez-windifferent If you explain your use case maybe we can find another solution.

blahetal:
Having a full blown implementation which would check sub/properties and functions of all detected identifiers would be great.

Issue #328, opened 2024-11-13
Broken localization


cpiber:
#318 broke localization

Some strings are still localized (e.g. `ErrorMessages.Format` is still used), but many actually useful messages are hardcoded again, like "Unknown identifier"
https://github.com/dynamicexpresso/DynamicExpresso/blob/590910db84145fcbc268258cc20739e931882c44/src/DynamicExpresso.Core/Exceptions/UnknownIdentifierException.cs#L10

davideicardi:
Thank you for the issue. You are right. We will try to resolve this issue in the future. As usual, any help is appreciated ;-) .

cpiber:
@metoule My PR that fixes it is already merged, I forgot to close this

PR #329, opened 2024-11-15
[#328] Introduce localized strings for new errors

cpiber:
This PR fixes the generation of `ErrorMessages.Designer.cs` (the change to the project file was not committed). It also introduces resource strings for all hardcoded strings returned to the user. I also added the German translations for those new strings.


davideicardi (approved):
Thank you!

Issue #325, opened 2024-11-07
Null check for property of parameter with late binding does not work anymore


AndreasKorn:
Checking a property of an input (parameter) object for null stopped working in 2.17.1.
This worked fine in 2.16.1.

```
  var options = InterpreterOptions.DefaultCaseInsensitive | InterpreterOptions.LambdaExpressions | InterpreterOptions.LateBindObject;
  var interpreter = new Interpreter(options);

  var input = new
  {
    //Prop1 = (object?)null, // works
    Prop1 = new { }, // does not matter what type
  };
  var expressionDelegate = interpreter.ParseAsDelegate<Func<object, bool>>($"input.Prop1 == null", "input")!;
//  var expressionDelegate = interpreter.ParseAsDelegate<Func<object, bool>>($"input.Prop1 == (object)null", "input")!; // also works
  var value = expressionDelegate(input); // exception

```

With 2.17.1 this throws an exception:
_System.InvalidCastException : Unable to cast object of type '<>f__AnonymousType2' to type 'System.Nullable`1[System.Boolean]'._

davideicardi:
Thank you for the bug report. We will try to investigate.

@metoule Do you have any idea of which change could have caused this behavior?

metoule:
I've tracked the change in behavior to #296. I'll try to investigate why it caused the new issue.

metoule:
I'll open a PR to revert the changes made in #296. Those changes force the conversion of a dynamic parameter to the function's parameter type. We should instead create a late binding expression. The fix will be more complex, so it will be in a distinct PR.

Issue #324, opened 2024-11-06
ArgumentException when calling Count() method from object returned by dynamic


alser:
- `DynamicExpresso.Core 2.17.1`
- .NET 8.0

Issue is reproduced in following default console template:

```xml
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net8.0</TargetFramework>
        <ImplicitUsings>enable</ImplicitUsings>
        <Nullable>enable</Nullable>
    </PropertyGroup>

    <ItemGroup>
      <PackageReference Include="DynamicExpresso.Core" Version="2.17.1" />
    </ItemGroup>

</Project>
```

```cs
using System.Dynamic;
using DynamicExpresso;

var interpreter = new Interpreter();
var expr = interpreter.Parse("p.Value.Count()", new Parameter("p", typeof(SampleParameter)));
var result = expr.Invoke(new SampleParameter());

Console.WriteLine($"Result: {result}");

public sealed class SampleClass
{
    public int Count() => 42;
}

public sealed class SampleParameter : DynamicObject
{
    public override bool TryGetMember(GetMemberBinder binder, out object? result)
    {
        switch (binder.Name)
        {
            case "Value":
                result = new SampleClass();
                return true;

            default:
                result = null;
                return false;
        }
    }
}
```

Here we have parameter `SampleParameter` which is a dynamic object. Its dynamic property `Value` returns some other non-dynamic object `SampleClass`, which in turn has method named `Count()`.

Trying to parse expression like `"p.Value.Count()"` results in exception:

```
System.ArgumentException: Type System.Collections.Generic.IEnumerable`1[TSource] contains generic parameters (Parameter 'type')
   at System.Dynamic.Utils.TypeUtils.ValidateType(Type type, String paramName, Int32 index)
   at System.Dynamic.Utils.TypeUtils.ValidateType(Type type, String paramName, Boolean allowByRef, Boolean allowPointer)
   at System.Linq.Expressions.Expression.Convert(Expression expression, Type type, MethodInfo method)
   at DynamicExpresso.Resolution.ExpressionUtils.PromoteExpression(Expression expr, Type type, Boolean exact)
   at DynamicExpresso.Resolution.MethodResolution.CheckIfMethodIsApplicableAndPrepareIt(MethodData method, Expression[] args)
[...]

PR #333, opened 2024-11-16
Revert changes introduced in #296

metoule:
The changes introduced in #296 caused #324 and #325, so this PR revert it. The proper fix will come in a later PR, because it will require creating a late binding expression, and it's a bit complex.

### DIFF
--- a/src/DynamicExpresso.Core/Resolution/ExpressionUtils.cs
+++ b/src/DynamicExpresso.Core/Resolution/ExpressionUtils.cs
@@ -36,7 +36,7 @@ namespace DynamicExpresso.Resolution
 					return Expression.Convert(expr, genericType);
 			}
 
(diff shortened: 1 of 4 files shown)


src/DynamicExpresso.Core/Resolution/ExpressionUtils.cs
```diff
@@ -40,13 +36,9 @@ public static Expression PromoteExpression(Expression expr, Type type, bool exac
 					return Expression.Convert(expr, genericType);
 			}
 
-			if (TypeUtils.IsCompatibleWith(expr.Type, type) || expr is DynamicExpression)
+			if (TypeUtils.IsCompatibleWith(expr.Type, type))
```
metoule, line 39: This is the actual fix, the rest is just cosmetic.

davideicardi (approved):

PR #342, opened 2025-02-08
Fix cast expression grammar

### LINKED ISSUE #341: (T)x precedence is wrong
The cast operator “(T)x” precedence is not correct.
```
Interpreter interpreter = new();
lambda = interpreter.Parse("(int)x == 1", new Parameter("x", typeof(int)));
lambda.Invoke(1)};
```     
The result of Invoke is **1** and  lambda,Expression value is **{Convert((x == 1), Int32)}**. According to operators precedence they should be **true** and **{Convert((x == 1), Int32)}** respectively.

metoule:
According [to the specs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1297-cast-expressions), a cast expression is:

```
cast_expression
    : '(' type ')' unary_expression
    ;
```

In the current implementation, the grammar is 

```
cast_expression
    : '(' type ')' expression
    ;
```

which is incorrect.

Fixes #341

### DIFF
--- a/test/DynamicExpresso.UnitTest/GithubIssues.cs
+++ b/test/DynamicExpresso.UnitTest/GithubIssues.cs
@@ -865,6 +865,17 @@ namespace DynamicExpresso.UnitTest
 			var expressionDelegate = interpreter.ParseAsDelegate<Func<object, bool>>($"input.Prop1 == null", "input");
 			Assert.IsFalse(expressionDelegate(input));
 		}
+
+		[Test]
+		public void GitHub_Issue_341()
+		{
+			var interpreter = new Interpreter();
+
+			var x = 1L;
+			interpreter.SetVariable("x", x);
+
+			Assert.AreEqual((int)x == 1, interpreter.Eval<bool>($"(int)x == 1"));
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


test/DynamicExpresso.UnitTest/GithubIssues.cs
```diff
@@ -865,6 +865,17 @@ public void GitHub_Issue_325()
...
+
+		[Test]
+		public void GitHub_Issue_341()
+		{
+			var interpreter = new Interpreter();
+
+			var x = 1L;
+			interpreter.SetVariable("x", x);
+
+			Assert.AreEqual((int)x == 1, interpreter.Eval<bool>($"(int)x == 1"));
```
davideicardi, line 877: Why the `$` string interpolation?

davideicardi (approved):
Thank you!

test/DynamicExpresso.UnitTest/GithubIssues.cs
metoule: You're right, it's not needed!

PR #345, opened 2025-02-09
NUnit: migrate to constraints model

metoule:
Starting with NUnit 4, the "classic" model of assertions is no longer available. This PR updates the test dependencies to the latest versions, including NUnit, and use the new [constraint model](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertion-models/constraint.html). 

This means that instead of writing:

```c#
Assert.AreEqual(23, target.Eval("myVar"));
```

we must write 

```c#
Assert.That(target.Eval("myVar"), Is.EqualTo(23));
```

Note that the goal is to have up-to-date dependencies, but we can stay on the 3.x version if you don't like the constraint model :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Variables can be used inside expressions with `Interpreter.SetVariable` method:
 ```csharp
 var target = new Interpreter().SetVariable("myVar", 23);
 
-Assert.AreEqual(23, target.Eval("myVar"));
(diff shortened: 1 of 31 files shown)


davideicardi (approved):
Thank you!

I admin that the new syntax looks a little bit "unusual" for me, but using legacy code/syntax is never a good idea. LGTM!

PR #346, opened 2025-02-26
Feature/custom member access

ANerdyElva:
We have a project that requires running expressions on mostly configuration based objects (in reality they're nested dictionaries with per-member/per-dict configuration of what variables are stored inside, with type information).
These expressions may in the future be written by people with limited technical experience, so `field * 5` is easy to explain but `(int)data["field"] * 5` is harder, and it gets worse from there (`(int)((List<Dictionary<string,object>>)this["child"])[0].field` vs `child.field`)

We could use dynamic objects but that would require generating a lot of them as it's a complex hierarchical data structure, and this would lose type information, as well as performance.

Instead I've made a small interface that serves as an extension point, if it's used then on attempting to resolve member access it will first attempt to call `bool TryGetMemberAccess(Expression leftHand, string identifier, out Expression result)`.
If it returns false normal code flow resumes, if true is returned result should be an Expression that represents the member access.

This can of course be anything, in the unit tests it generates a dictionary access and a cast.
Case insensitivity must be implemented by the implementer of IMemberAccessProvider if they with to use case insensitivity.

### DIFF
--- a/src/DynamicExpresso.Core/IMemberAccessProvider.cs
+++ b/src/DynamicExpresso.Core/IMemberAccessProvider.cs
@@ -1,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Text;
+
+namespace DynamicExpresso
+{
+    public interface IMemberAccessProvider
+    {
+		bool TryGetMemberAccess(Expression leftHand, string identifier, out Expression result);
+	}
+}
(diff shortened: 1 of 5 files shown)


metoule:
To be honest, that looks like a fairly niche requirement. We could imagine providing support for custom resolution, but we'd want to support more than member access: we'd probably want to support everything handled by the `MemberFinder`. 

Would it be possible for you to preprocess the string instead? If I understand your use case correctly, you could use `Detector.DetectIdentifiers` to find `field` and `child`, and replace them with a parameter with the expected value.

Issue #298, opened 2023-09-26
Parseasdelegate performance


acquleo:
I'm running an application which is parsing 100000 expressions ready to be evaluated at runtime.
The parsing of the expression using parseasdelegate takes forever.
Do you know any limitations of optimization to be made in order to speed up this process?
Thanks

metoule:
Do you have an example of an expression that your application evaluates? It could help pinpoint any bottleneck.
Can you also define what "forever" is to you? I benchmarked the parsing in the past, and it took about 10ms per expression if I remember correctly.

acquleo:
I'll give more details tomorrow when I'm at work.

If it would be 10ms for each parse then 100000 expressions would be 16 minutes of parsing so it will fit my forever sentence because I've stopped the application long before.
Using parse which returns a lambda is quicker but I've got an impressive memory usage.
2 GB of RAM with those 100000 expressions.

I'm trying to figure out the best compromise of memory usage and startup time.

All the expressions are like the following:
{Id1} | {id2} | {id3} | {id4}
Then the expression is elaborated by a regex replace to the following expression:
Al("id1") | Al("id2") | Al("id3") | Al("id4")

The execution time of the regex is very quick.

Then the resulting expression is parsed by dynamic Expresso, Al is a function that takes a string in input and returns a boolean.

Leonardo

davideicardi:
@acquleo  

If your expressions are all like the one you provided (with different values I suppose), probably DynamicExpresso is not the best tool for the job. Because you don't really need it for this kind of scenario. Maybe you can just use a regex to get all the values and do a custom processing.

Anyway in the past I had a similar scenario where I have a lof of expressions. I solved it using a "lazy" parsing approach: I don't parse all expressions in the initialization but only the ones that I use (like a parsing "on demand"), and then cache the result.

acquleo:
You're right.
I'm evaluating another expression evaluator which provides enough functionalities and better performance.

acquleo:
You're right.
I'm evaluating another expression evaluator which provides enough functionalities and better performance.
I'm using ncalc instead of dynamic Expresso.

metoule:
Hi @acquleo, 
thanks again for reporting this. Ncalc is indeed a lot faster than DynamicExpresso, but thanks to your example, I was able to improve performance (see #349 if you're interested in the details). We're still a lot slower, but not by as much :D

PR #349, opened 2025-03-15
⚡Improve performance

metoule:
This PR has been triggered by #298. I used [BenchmarkDotNet](https://benchmarkdotnet.org/) to compare DynamicExpresso and Ncalc for the kind of expression reported in #298:

```c#
using BenchmarkDotNet.Attributes;
using DynamicExpresso;
using NCalc;

namespace Benchmark;

public static class Program
{
	public static void Main()
	{
		BenchmarkDotNet.Running.BenchmarkRunner.Run<DynamicExpressoVsNCalcBenchmark>();
	}
}

public class DynamicExpressoVsNCalcBenchmark
{
	private string _expressionText;
	private Interpreter _dyn;

	[Params(5, 20)]
	public int N;

	[GlobalSetup]
	public void Setup()
	{
		_dyn = new Interpreter();
		_dyn.SetFunction("Al", (Func<string, bool>)(id => false));
		_expressionText = string.Join(" || ", Enumerable.Range(1, N).Select(i => $"Al(\"id{i}\")"));
	}

	[Benchmark]
	public System.Linq.Expressions.Expression<Func<bool>> DynamicExpressoParsingOnly()
	{
		return _dyn.ParseAsExpression<Func<bool>>(_expressionText);
	}

	[Benchmark]
	public bool DynamicExpressoEval()
	{
		return _dyn.Eval<bool>(_expressionText);
	}

	[Benchmark]
	public bool NcalcEval()
	{
		var expression = new Expression(_expressionText);
		expression.Functions["Al"] = (args) =>
		{
			return false;
		};

		return (bool)expression.Evaluate();
	}
}
```

DynamicExpresso always compiles the expression tree, while Ncalc uses an Expression visitor to evaluate the expression, and only compiles the expression tree if the user requires a lambda expression. Ncalc doesn't seem to support custom functions in that case, so to only consider the cost of the actual parsing in DynamicExpresso, I've added an additional benchmark that doesn't eval the expression. 

I've iterated multiple times: each commit reduces the execution time a bit. The best optim was the one that removes reflection to identify operators.

We're still a lot slower than Ncalc.

Here are the results.

### Baseline: before any optim

| Method                     | N  | Mean         | Error       | StdDev    | Median       |
|--------------------------- |--- |-------------:|------------:|----------:|-------------:|
| DynamicExpressoParsingOnly | 5  |  35,259.6 ns |   679.69 ns | 883.79 ns |  34,719.5 ns |
| DynamicExpressoEval        | 5  | 180,923.1 ns | 1,020.49 ns | 852.16 ns | 180,767.2 ns |
| NcalcEval                  | 5  |     656.7 ns |     1.81 ns |   1.60 ns |     656.8 ns |
| DynamicExpressoParsingOnly | 20 | 149,459.6 ns |   914.39 ns | 810.58 ns | 149,296.0 ns |
| DynamicExpressoEval        | 20 | 399,101.4 ns |   926.32 ns | 821.16 ns | 398,954.1 ns |
| NcalcEval                  | 20 |   2,698.5 ns |    10.08 ns |   9.42 ns |   2,701.2 ns |

### After all optimizations

| Method                     | N  | Mean         | Error       | StdDev      |
|--------------------------- |--- |-------------:|------------:|------------:|
| DynamicExpressoParsingOnly | 5  |  15,655.8 ns |   162.43 ns |   126.82 ns |
| DynamicExpressoEval        | 5  | 160,163.4 ns |   761.47 ns |   594.50 ns |
| NcalcEval                  | 5  |     640.0 ns |     9.03 ns |     9.66 ns |
| DynamicExpressoParsingOnly | 20 |  63,925.0 ns |   286.82 ns |   239.50 ns |
| DynamicExpressoEval        | 20 | 314,246.6 ns | 1,869.79 ns | 1,561.36 ns |
| NcalcEval                  | 20 |   2,798.6 ns |    12.05 ns |    10.06 ns |

For parsing only, for N=20, we went from 149us to 64us, or a 57% decrease.


holdenmai:
Performance commits, my favorite!

Apologies if I missed something in my suggestions as I am reviewing from a phone while I am having trouble sleeping but overall I think the changes are pretty solid. 

In the overload class, instead of having an actual Lazy instance instantiated every time, why not do the last yourself? Having a private MethodData? field that you set if needed. This would end up saving an object allocation, but not sure how much that is in the hot path of the benchmarking.

I'm not sure the exact code for it, but in the FindMethods I have a feeling that there is a way to reduce the number of lists created as the List is only needed in the case of ambiguous matches and since it looks like we don't actually use the List result in the exception, we could return something like a result struct with HasAmbiguousMatch and the BestMethod.  

This is specifically what I was mentioning when I said I may be missing something, but is there a reason that the MethodSignatures arrays are public and not some kind of ReadOnly?  From my phone review I'm not quite sure if those are truly public to external code, but if they are I don't think we would want any chance of those being modified. 

Also, if you are trying to get rid of Linq, I see .Skip calls eagerly awaiting removal :)

Overall I think the changes look solid and I like the resulting change to the processing.

metoule:
Thanks for the feedback :) 

> In the overload class, instead of having an actual Lazy instance instantiated every time, why not do the last yourself?

There's only one `Lazy` per overload, but there's only a handful of custom functions, so the number of allocations is not on the hot path. What's costly is the call to resolve the `Invoke` method, because that requires reflection. I'll still remove it, good catch 👍 

> Is there a reason that the MethodSignatures arrays are public and not some kind of ReadOnly?

The class itself is internal, so the members themselves can be marked public without being available outside of the assembly. We could indeed use some kind of read-only collections, but I'm not sure it would bring much. 

> if you are trying to get rid of Linq, I see .Skip calls eagerly awaiting removal :)

For now, I focused on Linq calls where the lambda captures variables: in that case, the C# compiler will create a class to hold the captured variable, and will instantiate that class whenever the method is called. If the lambda doesn't capture any variable, then there's only a single instantiation, so it's not too costly.
That being said, we could try to remove all calls to see if it improves even more :)

> in the FindMethods I have a feeling that there is a way to reduce the number of lists created

I'll have a look :)

holdenmai:
@metoule I finally had an opportunity to start looking at some of this in an actual development environment and one of the things I noticed is that it looks like the benchmark comparison isn't an apples to apples comparison.  Ncalc caches the compiled expression, which can easily be done by end users of DynamicExpresso.  Obviously that doesn't mean we shouldn't be looking into improving our parsing speed, but the picture may not be as bleak is it once seemed.

If you tell Ncalc to use NoCache that takes Ncalcs benchmarks to 13,450 ns and 52,359 ns for N = 5, 20.  So we definitely still have major room for improvement, however after your changes if a caching strategy was followed these two libraries are getting a lot closer.

holdenmai:
I've been consumed by this PR and have built in some additional changes around performance based on the changes you have already made.  I almost made it so that there isn't any list allocation in the FindMethod logic, but the single list that is allocated will only have more than 2 items in extremely rare cases whereas the old logic regularly could hit 10 or 11 that it then has to reprocess.

  My timings from the same benchmarking logic after improvements is


Method | N | Mean | Error | StdDev
-- | -- | -- | -- | --
DynamicExpressoParsingOnly | 5 | 10,246.8 ns | 200.6 ns | 281.22 ns
DynamicExpressoEval | 5 | 184830.1 ns | 3602.97 ns | 5167.27 ns
NcalcEval | 5 | 784.3 ns | 15.46 ns | 26.25 ns
NcalcEvalNoCache | 5 | 11,609.5 ns | 202.24 ns | 179.28 ns
DynamicExpressoParsingOnly | 20 | 63,925.0 ns | 286.82 ns | 239.50 ns
DynamicExpressoEval | 20 | 314,246.6 ns | 1,869.79 ns | 1,561.36 ns
NcalcEval | 20 | 2,798.6 ns | 12.05 ns | 10.06 ns
NcalcEvalNoCache | 20 | 44130.1 ns | 398.75 ns | 372.99 ns




Timings from before my changes
Method | N | Mean | Error | StdDev     
-- | -- | -- | -- | --
| DynamicExpressoParsingOnly | 5  |  13,778.0 ns |   223.15 ns |   197.81 ns
| DynamicExpressoEval        | 5  | 191,236.9 ns | 3,764.84 ns | 5,635.04 ns
| NcalcEval                  | 5  |     755.8 ns |    14.41 ns |    15.42 ns
| NcalcEvalNoCache           | 5  |  11,846.7 ns |   232.50 ns |   258.42 ns
| DynamicExpressoParsingOnly | 20 |  57,399.6 ns |   576.39 ns |   481.31 ns
| DynamicExpressoEval        | 20 | 334,871.9 ns | 5,342.14 ns | 4,997.04 ns
| NcalcEval                  | 20 |   3,027.6 ns |    57.66 ns |    61.70 ns
| NcalcEvalNoCache           | 20 |  43,733.4 ns |   585.24 ns |   518.80 ns




I'm not sure what's up with the DynamicExpressoParsingOnly N = 20 case since the Eval is proportionally faster.

Once this PR is merged in I can open up a new one with my changes.

PR #363, opened 2025-05-05
Code format: fix compilation suggestions

metoule:
* Use `var` everywhere (I can revert if not wanted)
* Remove unused method arguments
* Use pattern matching 
* Fix formatting issues


davideicardi (approved):
Thank you!

PR #365, opened 2025-05-15
No longer store MethodData as static variables

### LINKED ISSUE #361: Variable 'xxx' of type 'System.String' referenced from scope '', but it is not defined'
Hello!
After upgrading from 2.18 to 2.19 i got this weird error when evaulating an expression.
An example of the error stack:

Exception has occurred: CLR/System.InvalidOperationException
Exception thrown: 'System.InvalidOperationException' in System.Linq.Expressions.dll: 'variable 'FK_Country' of type 'System.String' referenced from scope '', but it is not defined'
   at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitParameter(ParameterExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitLambda[T](Expression`1 node)
   at System.Linq.Expressions.LambdaExpression.Compile()

ParameterExpression[0]
[0] [ParameterExpression] ={FK_Country}
CanReduce [bool] =false
DebugView [string] ="$FK_Country"
IsByRef [bool] =false
Name [string] ="FK_Country"
NodeType [ExpressionType] =Parameter
Type [Type] ={System.String}

The expressions is just a simple Eval<bool>( "FK_Country != \"France\"" )
Expression debug view : {(FK_Country != "France")}

Hope it can helps!

### LINKED ISSUE #364: Eval method is not thread safe and gives incorrect answers when executed in parallel
When doing evaluations in parallel in different threads it somehow seem that the values from the different evaluations conflict with each other and are used in unintended ways. I.e. variable value from one thread is used in another thread.

Issue is introduced in this change set
[Improve Performance (#349)](https://github.com/dynamicexpresso/DynamicExpresso/commit/aa23c9819758f4b0a5d70f9c59ee55bb50822c34)
in version 2.19.

And it can be reproduced with this unit test which passes in version 2.18 but not in 2.19.

```
using System.Collections.Generic;
using System.Globalization;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;
using NUnit.Framework;
namespace DynamicExpresso.UnitTest
{
 [TestFixture]
 public class ThreadingTest
 {
  private void DynamicExpresso_Interpreter_Eval1()
  {
   var parser = new Interpreter();
   parser.SetVariable("C0", 0.4731765);
   parser.SetVariable("x", 50);
   Assert.That(parser.Identifiers.First(i => i.Name == "C0").Expression.ToString(), Is.EqualTo("0.4731765"));
   Assert.That(parser.Identifiers.First(i => i.Name == "x").Expression.ToString(), Is.EqualTo("50"));
   var result = parser.Eval<double>("x*C0");
   Assert.That(result, Is.EqualTo(23.658825));
  }
  private void DynamicExpresso_Interpreter_Eval2()
  {
   var parser = new Interpreter();
   parser.SetVariable("C0", 0.4731765);
   parser.SetVariable("y", 1080);
   Assert.That(parser.Identifiers.First(i => i.Name == "C0").Expression.ToString(), Is.EqualTo("0.4731765"));
   Assert.That(parser.Identifiers.First(i => i.Name == "y").Expression.ToString(), Is.EqualTo("1080"));
   var result = parser.Eval<double>("y/C0");
   Assert.That(result, Is.EqualTo(2282.4464021353556));
  }
  private void DynamicExpresso_Interpreter_Eval_Sequence()
  {
   for (var i = 0; i < 10; i++)
   {
    DynamicExpresso_Interpreter_Eval1();
    DynamicExpresso_Interpreter_Eval2();
   }
  }
  [Test]
  public async Task DynamicExpresso_Interpreter_Eval_Threading()
  {
   Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
   const int noOfTasks = 5;
   var tasksToRun = new List<Task>();
   for (var i = 0; i < noOfTasks; i++)
   {
    var taskToRunLater = new Task(() => DynamicExpresso_Interpreter_Eval_Sequence());
    tasksToRun.Add(taskToRunLater);
   }
   foreach (var taskToRunLater in tasksToRun)
   {
    taskToRunLater.Start();
   }
   foreach (var taskToRunLater in tasksToRun)
   {
    await taskToRunLater;
   }
  }
 }
}
 

```

metoule:
#349 (via commit https://github.com/dynamicexpresso/DynamicExpresso/pull/349/commits/0346c9d0fa41ff0cbb0c2c5d2132a7bdd235fff0) removed the reflection code that was used to know the operands of binary and unary operators. Instead, it stored `MethodData` object as static objects, so that they could directly be used during the resolution process.

That proved to be an ill-advised choice: the resolution process mutates the `MethodData` class, but since they're stored as static variables, it was not thread-safe: thanks @mats-boisen for the investigation in #364!

This PR instead stores immutable objects as static variables. The `MethodData` used during resolution is therefore created for each resolution.

Fix #361 
Fix #364


holdenmai:
@metoule What are your thoughts on having the static fields and updated method signatures be Method base instead of the SimpleMethodSignature?  Obviously the actual ones you return would still be the simple class.

This may allow us to transform some things going forward if we make further changes/optimizations around how these operations are looked up.

davideicardi (approved):
Thanks!

davideicardi:
@metoule Thank you!

Just a question to fully understand:
When the `MethodData` will be created exactly with this new PR?

metoule:
> When the MethodData will be created exactly with this new PR?

In `Parser.PrepareOperandArguments`, when we call `MethodResolution.FindBestMethod(signatures, args)`: that overload converts the `MethodBase` to `MethodData`.

PR #369, opened 2025-09-06
Fix #368: generate proper expression for null-coalescing operators on nullable types

### LINKED ISSUE #368: Nullable variables (int?) cause error on evaluation when value is null
When setting a nullable variable (int?) with a null value, DynamicExpresso throws an evaluation error instead of handling the expression correctly.

```
var interpreter = new DynamicExpresso.Interpreter();

int? MyVar = null;

interpreter.SetVariable("MyValue", MyVar, typeof(int?));

var result = interpreter.Eval("(MyValue ?? 0) == 2");

```

Event These Evaluation Fails :

```
((int?)2 ?? 0) == 2      // Should be true, but throws
((int?)null ?? 0) == 2   // Should be false, but throws
```

**Expected behaviour :** The expression should evaluate without error and return false.

**Actual behavior:**  It throws an InvalidOperationException.

**Expected issue :**  DynamicExpresso treats (nullable ?? value) as returning object instead of the underlying type (int in this case).

Can someone please help me identify is it a bug/my wrong of of using and some other way possible to handle the scenario.

I am using Version : 2.19.2     (Latest as of now)

metoule:
The expression we generate for the null-coalescing operator doesn't work for nullable types, because there's no comparison operator between a nullable type and an object:

```c#
const object nullConst = null;
int? MyVar = null;
return MyVar == nullConst ? 0 : MyVar; // <-- doesn't compile
```

This PR changes the generated expression for nullable types from `expr == null ? exprRight : expr` to `expr.HasValue ? expr.Value : exprRight`

Fix #368

### DIFF
--- a/src/DynamicExpresso.Core/Parsing/Parser.cs
+++ b/src/DynamicExpresso.Core/Parsing/Parser.cs
@@ -249,7 +249,16 @@ namespace DynamicExpresso.Parsing
 			{
 				NextToken();
 				var exprRight = ParseExpressionSegment();
-				expr = GenerateConditional(GenerateEqual(expr, ParserConstants.NullLiteralExpression), exprRight, expr, errorPos);
+				if (TypeUtils.IsNullableType(expr.Type))
+				{
+					// expr.HasValue ? expr.Value : exprRight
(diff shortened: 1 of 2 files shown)


metoule:
It's actually ready, the other binary operators are properly resolved (e.g. `int? > int` works as intended).

davideicardi (approved):

PR #371, opened 2026-01-31
Optimize Lambda.Invoke / Interpreter.Eval performance + add Benchmarks

WAcry:
## Stack

- **This PR:** [#371](https://github.com/dynamicexpresso/DynamicExpresso/pull/371) -> **Follow-up PR** [#372](https://github.com/dynamicexpresso/DynamicExpresso/pull/372) 
## What & why

This PR reduces overhead in hot invocation/evaluation paths by:
- moving per-lambda computed data into a dedicated context object (so we compute once, reuse many times),
- removing avoidable work from the `Invoke(object[])` hot path, and
- avoiding IL generation when evaluating one-off expressions where compilation cost dominates.

It also adds a BenchmarkDotNet project so we can measure both **time** and **allocations** going forward.

## Changes

- **Centralize invocation work in `InvocationContext` (internal):**
  - Cache the compiled delegate for reuse.
  - Precompute an index mapping from **declared parameters → used parameters**, so invocations don’t need to repeatedly match by name.

- **Optimize `Lambda.Invoke(object[])`:**
  - In the hot path, build the used-arguments array directly in declared order (avoids name matching / dictionary-like lookups).

- **`Interpreter.Eval(...)` for one-off expressions:**
  - Prefer interpreted execution to avoid paying IL generation overhead when the expression is evaluated once.

- **Add benchmarks:**
  - New `BenchmarkDotNet` project: `DynamicExpresso.Benchmarks`
  - Measures `invoke` / `eval` time and allocations.

## Benchmark results

Environment:

- BenchmarkDotNet `v0.14.0`
- Windows 11 (`10.0.26200.7705`)
- 11th Gen Intel Core i7-11800H (16 logical / 8 physical)
- .NET `8.0.23` (x64 RyuJIT)
- .NET SDK `10.0.102`

### Comparison (master vs PR #371)

| Method | master Mean (ms) | PR #371 Mean (ms) | Speedup | master Alloc (MB) | PR #371 Alloc (MB) | Alloc reduction |
|---|---:|---:|---:|---:|---:|---:|
| Invoke cached lambda (object[]) | 221.70 | 10.04 | 22.08x | 260.93 | 9.16 | 28.49x |
| Invoke cached lambda (IEnumerable&lt;Parameter&gt;) | 175.80 | 96.20 | 1.83x | 191.50 | 37.38 | 5.12x |
| Eval (IEnumerable&lt;Parameter&gt;) | 25503.70 | 16346.04 | 1.56x | 17172.52 | 16800.18 | 1.02x |

Notes:
- The largest improvement is **`Invoke(object[])`**: ~**22x faster**, ~**28x fewer allocations** (in this benchmark scenario).
- `Eval(...)` is still much slower than a cached lambda, but improves by ~**1.56x** here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,6 +109,7 @@ stylecop.*
 ~$*
 *.dbmdl
 Generated_Code #added for RIA/Silverlight projects
+BenchmarkDotNet.Artifacts/
 
 # Backup & report files from converting an old project file to a newer
 # Visual Studio version. Backup files are not needed, because we have git ;-)
--- a/DynamicExpresso.sln
+++ b/DynamicExpresso.sln
@@ -12,23 +12,62 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{4088369E-A
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DynamicExpresso.Core", "src\DynamicExpresso.Core\DynamicExpresso.Core.csproj", "{C6B7C0D2-B84A-4307-9C61-D95613DB564D}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "benchmark", "benchmark", "{09EED85C-BE3C-7566-DC0E-2E8E43466740}"
+EndProject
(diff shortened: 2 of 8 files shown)


davideicardi:
Closed for inactivity

PR #377, opened 2026-07-13
docs: add TagBites.Expressions to the list of similar projects

tomaszmalik:
I'm the author of TagBites.Expressions - https://github.com/TagBites/TagBites.Expressions

I used DynamicExpresso before and it worked well for me, so thanks for maintaining it! I needed to support a larger part of C# syntax than DynamicExpresso's do, so I used Roslyn to skip parsing and focus only on converting syntax tree to expression tree. This project is also on MIT license, so I hope someone find it helpful.


davideicardi:
Yes sure! It looks a very good work! Congrats!

davideicardi (approved):

